PR #61, opened 2014-09-05
journal: Connection Read/Write Deadlines

fujin:
* Implement hard-coded 10msec read/write connection deadline.
* Journal i/o socket write throughput responsiveness suffers greatly. Writes
  can be impacted. We see log drops at the rate of about 1/500.


philips:
I would rather not add workarounds to the library when they are fixed in the kernel. I am going to close this unless someone has a really strong opinion.

PR #72, opened 2015-01-25
*: switch to line comments for copyright

jonboulle:
Build tags are not compatible with block comments. Also adds copyright header
to a few places it was missing.

/cc @bcwaldon


bcwaldon:
The copyright year shouldn't be modified as the source code the copyright protects has not changed at all. I understand that it's more complicated to preserve the original year, though, so LGTM.

PR #82, opened 2015-04-09
fix up test script and vendor godbus as required by the test script

mischief:
this is quite unfortunate, but go-systemd exposes github.com/godbus/dbus types in its api. this complicates its use by other go programs if they also use github.com/godbus/dbus, and can cause catastrophe if some other go program or a library it depends on *also* uses github.com/godbus/dbus..

but at least the tests can run. :-)

cc @bcwaldon @jonboulle @philips


bcwaldon:
@mischief this is fine by me - just update the commit message to follow CoreOS guidelines

bcwaldon:
Actually, given that this is a library, we should not be rewriting the imports...right?

jonboulle:
I don't think we reached a consensus on that...

crawford:
Why wouldn't you want to vendor dependencies? Is there any other way to guarantee correct behavior?

jonboulle:
@crawford https://github.com/tools/godep/issues/109

mischief:
if the dependencies aren't vendored, i can't see how the eng repo's build/test script can work, since the dependencies will live outside of the `GOPATH=${PWD}/gopath` set by the build script.

should just add `go fmt` and `go vet` to the test script here and not fiddle with `GOPATH`?

jonboulle:
The alternative/compromise would be to `go get` the dependent packages in the build script after setting up the custom GOPATH.

crawford:
But then the value of the test is lost. Depending on when the test is run, it may pass or fail.

jonboulle:
Right, which is the same case as today.
Am 09.04.2015 13:19 schrieb "Alex Crawford" <notifications@github.com>:

marineam:
Just depend on usual `go get` and `go test` for libraries. If you want to construct a test script that can be used in a repeatable way in a CI system you can vendor the deps and set GOPATH appropriately. Expecting downstream users of this library to use the exact copy of godbus as vendored here is unworkable, a user of this library may in fact have multiple users of godbus, they all cannot have their own copy.

It is on the final project (where 'main' is) to get dependencies right and ensure all the libraries work together. Unless go's vendoring tooling dramatically improves never ever rewrite import paths in general purpose libraries or everyone will hate you. <3

PR #85, opened 2015-05-04
activation: allocate slice if we know the size

tgulacsi:
Use for range, and use the knowledge the everything is zeroed on
allocation, so gaps will be nil in the slice.


bcwaldon:
@tgulacsi Thanks for the contribution!

PR #86, opened 2015-05-17
Add gatewayd journal client

chancez:


philips:
lgtm, could you add a MAINTAINERS file in this directory with your name?

jonboulle:
Copyright header please.

ecnahc515:
@jakedt @philips Done.

gatewayd/journal_test.go
```diff
@@ -0,0 +1,167 @@
...
+	"reflect"
+	"testing"
+)
+
+func TestGetEntries(t *testing.T) {
+
+	var expected = []*Entry{
+		// currently any fields which systemd prepends with "_" or "__" will be hidden
+		&Entry{
+			CURSOR:                  "s=c60ae6415a0e4531a2064c8cdb11a0f4;i=587;b=d5f050252af342a2817795b375889f78;m=b11f9d6;t=5120f62678669;x=74d4678ad474a6b6",
```
jonboulle: These don't match to the actual names

gatewayd/journal_test.go
```diff
@@ -0,0 +1,167 @@
...
+		entries = append(entries, entry)
+	}
+
+	if len(entries) != 2 {
+		t.Errorf("Expected 2 entries and found %d", len(entries))
+		return
+	}
+
+	var i int
+	for entry := range entries {
```
jonboulle: for i, entry := range entries ?

gatewayd/entries.go
```diff
@@ -0,0 +1,138 @@
...
+		var entry *Entry
+		if err = decoder.Decode(&entry); err == io.EOF {
+			err = nil
+			return
+		} else if err != nil {
+			return
+		}
+		entries = append(entries, entry)
+	}
+	return
```
jonboulle: unreachable, I think go is smart enough to not complain about this any more?

gatewayd/entries.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		return
+	}
+	entries, err = readerToEntries(resp.Body)
+	return
+}
+
+func (c *Client) Follow(opts *Options) (*Stream, error) {
+	if opts == nil {
+		opts = &Options{Follow: true}
```
jonboulle: perhaps just &Options{} here?

gatewayd/journal_test.go
ecnahc515: That's weird..I must have accidentally broken the tests in a rebase before I submitted the PR. Should be fixed. I also found a bug with something else, noted in a comment to this PR.

ecnahc515:
I found a small bug in my implementation with the Message field. Things were not marshalling correctly into it. I found a slightly better alternative to a getter method as well. I basically implement a custom Unmarshal function that handles the varying types of data from the journald `Message` field and store that data as a string on the struct's `Message` Field.

gatewayd/journal.go
```diff
@@ -0,0 +1,123 @@
...
+		return e, err
+	}
+	switch t := e.MessageRaw.(type) {
+	case []byte:
+		e.Message = string(t)
+	case string:
+		e.Message = t
+	// When systemd encodes to JSON, if it finds any non-printable or non-UTF8
+	// character it gets serialized to a number array.
+	case []interface{}:
```
ecnahc515: So this is basically something I adapted from code I used elsewhere previously, with some changes. When systemd is serializing a field that contains non-printable or non-UTF8 is serialized as a number array instead. This is necessary to handle binary data. Since `MessageRaw` still contains this, I purely have this so it will actually store a form of string representation in case a client wants to display the message.

gatewayd/journal.go
```diff
@@ -0,0 +1,121 @@
...
+func setRangeHeader(req *http.Request, rangeOpts RangeOptions) {
+	header := fmt.Sprintf("entries=%s:%d:%d", rangeOpts.Cursor, rangeOpts.NumSkip, rangeOpts.NumEntries)
+	req.Header.Set("Range", header)
+}
+
+type decoder interface {
+	Decode(v interface{}) error
+}
+
+func DecodeEntry(dec decoder, r io.Reader) (*Entry, error) {
```
ecnahc515: I adapted this from somewhere else to handle storing a string representation on the `Entry` struct.

http://www.freedesktop.org/wiki/Software/systemd/json/ has details about it. The `[]interface{} case` should deal with non-printable non-utf8 characters. But I'm not 100% sure that's correct

lucab:
@chancez I randomly stepped on this today. Do we still want to drive this toward master or should I just close this PR?

chancez:
@lucab It can be closed, but I also see no reason it couldn't be merged.

lucab:
@chancez I seem to understand the original usecase is not anymore pressing. While I don't have any particular issue with this PR, I see this is mostly just an HTTP API client not strictly related to libsystemd. I'm closing this for now, but I think it can be easily resurrected or split in its own projects if there is any need for this.

PR #87, opened 2015-05-21
util: add GetRunningSlice, RunningFromSystemService, CurrentUnitName

jonboulle:
This adds two utility functions which can be used by processes to determine
their context (slice), and whether they are running from a systemd "system"
service.

This is a port of [code in rkt](https://github.com/coreos/rkt/blob/d23669f63644f63c1c5b61ea3c95f90ee9ab66f9/stage1/init/init.go#L399), with a few improvements (correctly freeing CStrings, broader portability for older systemd versions).

To check for whether we're running from a "system service" we use the approach suggested by Lennart (and described [here](https://github.com/coreos/rkt/issues/911#issuecomment-103853978)):

	hence: to detect if something is a system service, the best approach is
	to check if it is not a user process. There are two kinds of user
	processes: those running in session scopes, and those running of the
	per-user systemd --user instance (which is a per-user singleton shared
	by every session of the user). sd_pid_get_owner_uid() can recognize
	session scopes and the systemd --user service, extracts the uid from it
	and returns it. It's hence suitable as a check if something is user
	code or not.

Unfortunately this results in false positives: on CoreOS Linux, where there's no PAM, logging in does not result in a user slice/scope being created, and instead the session runs under `systemd-sshd.slice`:

	core@coreos_production_qemu-633-1-0 ~ $ systemd-cgls |fgrep $PPID -B3
	  ├─system-sshd.slice
	  │ ├─sshd@9-10.0.2.15:22-10.0.2.2:44781.service
	  │ │ ├─18061 sshd: core [priv]
	  │ │ ├─18063 sshd: core@pts/2
	  │ │ ├─18064 -bash
	  │ │ ├─19669 systemd-cgls
	  │ │ └─19670 grep -F 18063 -B3

	core@coreos_production_qemu-633-1-0 ~ $ ./sysd
	slice = system-sshd.slice
	runningFromSystemService = true

compare to a Fedora 20 system where a user scope is created:

	jon@fedora20 ~ $ systemd-cgls | fgrep $PPID -B2
	  │ └─session-452.scope
	  │   ├─2809 sshd: jon [priv]
	  │   ├─2814 sshd: jon@pts/13
	  │   ├─2816 -bash
	  │   ├─2928 systemd-cgls
	  │   └─2929 fgrep --color=auto 2814 -B2
	
	jon@fedora20 ~ $ ./sysd 
	slice = user-1000.slice
	runningFromSystemService = false

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -34,6 +34,15 @@ package util
 // }
 //
 // int
+// my_sd_pid_get_unit(void *f, pid_t pid, char **unit)
+// {
+//   int (*sd_pid_get_unit)(pid_t, char **);
+//
+//   sd_pid_get_unit = (int (*)(pid_t, char **))f;
+//   return sd_pid_get_unit(pid, unit);
+// }
+//
+// int
 // my_sd_pid_get_slice(void *f, pid_t pid, char **slice)
 // {
 //   int (*sd_pid_get_slice)(pid_t, char **);
@@ -191,6 +200,44 @@ func RunningFromSystemService() (ret bool, err error) {
 	default:
 		err = fmt.Errorf("error calling sd_pid_get_owner_uid: %v", syscall.Errno(-errno))
 	}
+	return
+}
+
+// CurrentUnitName attempts to retrieve the name of the systemd system unit
+// from which the calling process has been invoked. It wraps the systemd
+// `sd_pid_get_unit` call, with the same caveat: for processes not part of a
+// systemd system unit, this function will return an error.
+func CurrentUnitName() (unit string, err error) {
+	var h *libHandle
+	h, err = getHandle()
+	if err != nil {
+		return
+	}
+	defer func() {
+		if err1 := h.Close(); err1 != nil {
+			err = err1
+		}
+	}()
+
+	sym := C.CString("sd_pid_get_unit")
+	defer C.free(unsafe.Pointer(sym))
+	sd_pid_get_unit := C.dlsym(h.handle, sym)
+	if sd_pid_get_unit == nil {
+		err = fmt.Errorf("error resolving sd_pid_get_unit function")
+		return
+	}
+
+	var s string
+	u := C.CString(s)
+	defer C.free(unsafe.Pointer(u))
+
+	ret := C.my_sd_pid_get_unit(sd_pid_get_unit, 0, &u)
+	if ret < 0 {
+		err = fmt.Errorf("error calling sd_pid_get_unit: %v", syscall.Errno(-ret))
+		return
+	}
+
+	unit = C.GoString(u)
 	return
 }
 
--- a/util/util.go
+++ b/util/util.go
@@ -43,17 +43,31 @@ package util
 //
 import "C"
 import (
+	"errors"
 	"fmt"
 	"os"
 	"syscall"
 	"unsafe"
 )
 
+// libHandle represents an open handle to the systemd C library
+type libHandle struct {
+	handle  unsafe.Pointer
+	libname string
+}
+
+func (h *libHandle) Close() error {
+	if r := C.dlclose(h.handle); r != 0 {
+		return fmt.Errorf("error closing %v: %d", h.libname, r)
+	}
+	return nil
+}
+
 // getHandle tries to get a handle to a systemd library (.so), attempting to
 // access it by several different names and returning the first that is
 // successfully opened. Callers are responsible for closing the handler.
-// Returns nil if no library can be found.
-func getHandle() (unsafe.Pointer, string) {
+// If no library can be successfully opened, an error is returned.
+func getHandle() (*libHandle, error) {
 	for _, name := range []string{
 		// systemd < 209
 		"libsystemd-login.so",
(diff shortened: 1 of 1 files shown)


jonboulle:
similarly, output from [steveej's test program](https://github.com/steveeJ/hello_go/blob/master/misc/systemd_detect.go):
```
core@coreos_production_qemu-633-1-0 ~ $ ./systemd_detect     
sd_pid_get_owner_uid: no such file or directory, 0
sd_pid_get_unit: errno 0, sshd@15-10.0.2.15:22-10.0.2.2:49468.service
sd_pid_get_user_unit: no such file or directory, 
sd_pid_get_session: no such file or directory, 
sd_pid_get_slice: errno 0, system-sshd.slice
inside unit? true
```

jonboulle:
```
15:53 <jonboulle> poettering: still struggling with using sd_pid_get_owner_uid to detect whether we're in a system service - CoreOS Linux doesn't have pam or any user slices/scopes, so it only ever results in ENXIO (nee ENOENT)
15:53 <jonboulle> https://github.com/coreos/go-systemd/pull/87
15:55 <•poettering> jonboulle: oh, well, yeah
15:55 <jonboulle> well yeah, what can we do instead? :-)
15:55 <•poettering> jonboulle: if you don't have PAM then login will be a system service like any oter
15:56 <•poettering> jonboulle: nothing
15:56 <jonboulle> urgh
15:56 <•poettering> pam_systemd is how sessions are registered
15:56 <•poettering> if you don't call into that then there are no sessions
15:56 <•poettering> quite frankly, it's pretty stupid to avoid PAM
15:56 <•poettering> i mean, it's fine to avoid it in an embedded device
15:56 <•poettering> but otherwise, for anything more generic it's really pointless not to use it
15:57 <•poettering> it provides you with a hook into session management
15:57 <•poettering> and if you don't have it systemd cannot hook into anything there...
15:57 <•poettering> the whole concept of users and sessions in systemd is bound to PAM
15:57 <•poettering> that all falls apart if you don't have PAM
```

philips:
@marineam what are your thoughts on PAM these days?

marineam:
Hey, we avoided a SSH vulnerability by avoiding PAM ;-)

We do need to enable PAM at some point though.

util/util.go
```diff
@@ -13,13 +13,183 @@
...
+	case serrno == syscall.ENOENT, serrno == syscall.ENXIO:
+		// Since the implementation of sessions in systemd relies on
+		// the `pam_systemd` module, using the sd_pid_get_owner_uid
+		// heuristic alone can result in false positives if that module
+		// (or PAM itself) is not present or properly configured on the
+		// system. As such, we also check if we're the session leader,
+		// which should be the case if we're invoked from a unit file,
+		// but not if e.g. we're invoked from the command line from a
+		// user's login session
+		ret = C.am_session_leader() == 1
```
iaguis: We already say it's an heuristic so it won't be perfect but Should we mention https://github.com/coreos/rkt/pull/1694#issuecomment-152166269 here?

jonboulle: https://github.com/coreos/go-systemd/commit/2a36382d7303f2868c56ebc94def3d75c12d36b1#diff-99f38c7c07bfcc03a528180716bafab5R142

iaguis: :+1:

jonboulle:
@vcaputo review?

iaguis:
In rkt we also use [`sd_pid_get_unit`](https://github.com/coreos/rkt/blob/master/stage1/init/init.go#L35). Should we add it here too?

jonboulle:
@iaguis https://github.com/jonboulle/go-systemd/commit/53b918d2845c80f8eb0b11f51b13ab7ee41a7470

util/util.go
```diff
@@ -13,13 +13,225 @@
...
+	if sd_pid_get_unit == nil {
+		err = fmt.Errorf("error resolving sd_pid_get_unit function")
+		return
+	}
+
+	var s string
+	u := C.CString(s)
+	defer C.free(unsafe.Pointer(u))
+
+	ret := C.my_sd_pid_get_unit(sd_pid_get_unit, 0, &u)
```
iaguis: This is not defined

jonboulle: hugrgghh

updated

util/util.go
```diff
@@ -13,13 +13,234 @@
...
+	}
+	defer func() {
+		if err1 := h.Close(); err1 != nil {
+			err = err1
+		}
+	}()
+
+	sym := C.CString("sd_pid_get_unit")
+	defer C.free(unsafe.Pointer(sym))
+	sd_pid_get_unit := C.dlsym(handle, sym)
```
iaguis: h.handle

jonboulle: three hours sleep last night :-(

jonboulle:
@iaguis let's go again!

iaguis:
LGTM

I tested it in rkt and seems to work :)

iaguis:
We were thinking about https://github.com/coreos/rkt/issues/1879 and what we should do for the next rkt release.

The idea was to print a warning instead of erroring out if systemd is running but we can't find the libsystemd shared library. In this way, we don't break rkt for NixOS and we'll try to find a proper fix for the next release.

Perphaps we can add a typed error for the case we can't find libsystemd so rkt can detect that and ignore it for now? What do you think?

@jonboulle

iaguis:
https://github.com/coreos/go-systemd/pull/127

PR #89, opened 2015-05-27
util: add IsRunningSystemd (similar to sd_booted)

### LINKED ISSUE #88: There should be a isSystemd helper function
There should be a isSystemd helper function to help services determine if they are running under systemd. This will aid services in support more systemd features such as sdnotify, without breaking ignoring errors when not running on systemd systems.

jonboulle:
fixes #88


iaguis:
LGTM

PR #91, opened 2015-06-10
unit: add NewUnitOption convenience function

iaguis:
This convenience function is used a lot in [rkt/stage1/init/pod.go](https://github.com/coreos/rkt/blob/master/stage1/init/pod.go) and I'd like to reuse it somewhere else.


jonboulle:
lgtm

PR #92, opened 2015-06-23
dbus: add NewSystemdConnection() for connecting directly to systemd.

vcaputo:
Systemd exposes a dbus-like interface at /run/systemd/private which is
just dbus without the auth/hello.

Using this connection Go programs may be written which utilize systemd's
dbus integration without requiring a dbus daemon.

### DIFF
--- a/dbus/dbus.go
+++ b/dbus/dbus.go
@@ -86,14 +86,28 @@ type Conn struct {
 // New establishes a connection to the system bus and authenticates.
 // Callers should call Close() when done with the connection.
 func New() (*Conn, error) {
-	return newConnection(dbus.SystemBusPrivate)
+	return newConnection(func() (*dbus.Conn, error) {
+		return dbusAuthConnection(dbus.SystemBusPrivate)
+	})
 }
 
 // NewUserConnection establishes a connection to the session bus and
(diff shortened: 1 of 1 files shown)


dbus/dbus.go
```diff
@@ -132,12 +141,17 @@ func newConnection(createBus func() (*dbus.Conn, error)) (*Conn, error) {
 	return c, nil
 }
 
-func dbusConnection(createBus func() (*dbus.Conn, error)) (*dbus.Conn, error) {
+func dbusConnection(skipAuthHello bool, createBus func() (*dbus.Conn, error)) (*dbus.Conn, error) {
```
crawford: I don't really like boolean args since they aren't really self-documenting. Would it be unreasonable to have:

```go
func dbusConnection(createBus func() (*dbus.Conn, error)) (*dbus.Conn, error) {
    // stuff
}
func dbusAuthedConnection(createBus func() (*dbus.Conn, error)) (*dbus.Conn, error) {
    dbusConnection(createBus)
    // auth
    // hello
}
```

vcaputo: I feel the same way about booleans, I'll fix it.

crawford:
LGTM

vcaputo:
woot

PR #95, opened 2015-06-25
dbus: add Auth step to NewSystemdConnection()

vcaputo:
Just the Hello needs to be omitted for direct to systemd connections,
oops!


crawford:
LGTM

Issue #98, opened 2015-07-29
go-systemd is broken at head


jcgregorio:
Looks like an upstream dependency changed:

$ go get github.com/coreos/go-systemd/dbus
# github.com/coreos/go-systemd/dbus
src/github.com/coreos/go-systemd/dbus/dbus.go:186: cannot use conn.Object("org.freedesktop.systemd1", dbus.ObjectPath("/org/freedesktop/systemd1")) (type dbus.BusObject) as type *dbus.Object in return argument: need type assertion

hinnerk:
This patch works for me on OS X 10.10.4, Go version 1.4.2.

pquerna:
fixed by #100

PR #99, opened 2015-07-29
Fix breakage from godbus changing *dbus.Object to dbus.BusObject interface.

jcgregorio:
Addresses issue #98.


bhundven:
This change fixes the build issue for me.
go version go1.3.3 linux/amd64

mikkeloscar:
This also needs to be updated in `machine1/dbus.go` and `login1/dbus.go`. I would send a PR but it might make more sense to include it here.

pquerna:
can be closed, fixed by #100

PR #100, opened 2015-08-15
Fix dbus api

pquerna:
Includes fix from #99 and also fixes login1 and machine1.


crawford:
LGTM. Can you prefix the commit messages with a scope (e.g. "dbus: ")?

pquerna:
@crawford updated.

PR #101, opened 2015-09-01
unit: add check for long lines

### LINKED ISSUE #69: Fail unit option parsing if line longer than 2000 characters
We originally discovered in fleet [0] that systemd silently truncates lines at 2000 characters. This has been confirmed upstream [1]. Unfortunately, I believe we need to add this check to our unit file parser so we can give a reasonable experience to users rather than making them chase down the same rabbit hole that I did.

[0] https://github.com/coreos/fleet/issues/992
[1] https://bugs.freedesktop.org/show_bug.cgi?id=85308

jonboulle:
fixes #69


jonboulle:
@crawford @vcaputo

vcaputo:
lgtm, but I'm not too familiar with the lexer implementation.

PR #103, opened 2015-09-03
[RFC] unit: pass through line continuations

jonboulle:
If a unit file contains line continuations, unit.Deserialize has been
stripping them out. Unfortunately this means it is impossible for
go-systemd users to work around the line length limit in systemd (1) by
splitting a long line into multiple ones, because go-systemd will just
rejoin them into a single, unusable line.

This changes unit.Deserialize to no longer strip out line continuations but
instead preserve them in the relevant options' values

Partial solution for coreos/fleet#992, along with #101

RFC - there are a few ways to approach this, none of which I'm particularly
happy with (in hindsight I think the lexer is a bad idea) - this is an attempt
at a minimally invasive solution.

### DIFF
--- a/unit/deserialize_test.go
+++ b/unit/deserialize_test.go
@@ -80,14 +80,15 @@ Environment= "FOO=BAR" "BAZ=QUX"
 			},
 		},
 
-		// line continuations respected
+		// line continuations unmodified
 		{
 			[]byte(`[Unit]
 Description= Unnecessarily wrapped \
     words here
 `),
 			[]*UnitOption{
-				&UnitOption{"Unit", "Description", "Unnecessarily wrapped      words here"},
+				&UnitOption{"Unit", "Description", `Unnecessarily wrapped \
+    words here`},
 			},
 		},
 
@@ -119,8 +120,8 @@ Description=Bar\
 Baz
 `),
 			[]*UnitOption{
-				&UnitOption{"Unit", "Description", "Bar # comment alpha"},
-				&UnitOption{"Unit", "Description", "Bar # comment bravo  Baz"},
+				&UnitOption{"Unit", "Description", "Bar\\\n# comment alpha"},
+				&UnitOption{"Unit", "Description", "Bar\\\n# comment bravo \\\nBaz"},
 			},
 		},
 
@@ -180,7 +181,7 @@ Description=Bar`),
 			[]byte(`[Unit]
 Description=Bar \`),
 			[]*UnitOption{
-				&UnitOption{"Unit", "Description", "Bar"},
+				&UnitOption{"Unit", "Description", "Bar \\"},
 			},
 		},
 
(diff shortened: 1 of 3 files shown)


jonboulle:
Could also do with more test cases.

Technically backwards-incompatible, although in practise I doubt it will be too problematic (famous last words)

vcaputo:
lgtm

jonboulle:
@vcaputo no comment!?

ZeissS:
Afaik we could just insert `\n` in the serialization step if the line becomes too long at arbitary places. 

That would obviously lead to slightly different unitfiles compared to the original ones.

jonboulle:
@ZeissS right, my goal with this version of the solution was to minimise
the modifications that we make to the original unit files so as to preserve
the "expected" experience

vcaputo:
Nope, no comment.

unit/deserialize_test.go
```diff
@@ -218,7 +219,7 @@ Description= words here `),
 Description= words here \
   `),
 			[]*UnitOption{
-				&UnitOption{"Unit", "Description", "words here"},
+				&UnitOption{"Unit", "Description", "words here \\"},
```
ZeissS: not sure about this test. when we serialize this again, wouldn't this lead to the next UnitOption being merged with this one by Systemd?

jonboulle: Sorry I am not sure I follow, would you mind elaborating?

crawford:
Looks adequate to me

unit/deserialize_test.go
ZeissS: This test shows that when you have a `\` at the end of the line followed by an empty line, that the empty line is dropped when building the `UnitOption` object.

Assuming now, that after this UnitOption are more UnitOption's, they are serialized directly after each other, right?

Initial Input
```
[Unit]
# The empty line afterwards makes this a file with 2 options
Description=long line \

Requires=docker.socket

```

```
[]UnitOptions{
  UnitOption{"Unit", "Description", "Something line with a backslash \\"},
  UnitOption{"Unit", "Required", "docker.service"},
}
```

Serializing should result into this file:

```
[Unit]
# This unitfile now has only one option
Description=long line \
Requires=docker.socket
```

ZeissS: For the `end_to_end_test.go`, you would add the following test:

```
{
			`[Unit]
Description=Demo \

Requires=docker.service
`,
			`[Unit]
Description=Demo \
Requires=docker.service
`,
		},
```

This tests is the current state, but this is IMHO broken. I would expect that it keeps the empty line (or drop the \ in this case).

jonboulle: Gotcha, I agree with your diagnosis

jonboulle:
closing in favour of #108

PR #104, opened 2015-09-04
Typo fix

dnaeon:


jonboulle:
I might push back on this as [iff is often intentional](https://en.wikipedia.org/wiki/If_and_only_if) but I can't see a strong reason for it in this case, so lgtm.

PR #106, opened 2015-09-12
*: expand/enhance test script

jonboulle:
Borrowing from the other common CoreOS test scripts, we now:
- build a self-contained gopath for testing if GOPATH is not set
- run all unit tests
- run actual systemd dbus tests if invoked as root
- run gofmt and go vet across all go source files

### DIFF
--- a/test
+++ b/test
@@ -1,3 +1,73 @@
-#!/bin/sh -e
+#!/bin/bash -e
+#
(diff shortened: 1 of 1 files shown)


jonboulle:
Shoulder LGTMed by @crawford

PR #107, opened 2015-09-14
[RFC] add journal bindings around sd-journal

jonboulle:
Work in progress. Based on Dan Mace's initial prototype announced here:
http://lists.freedesktop.org/archives/systemd-devel/2014-June/019840.html

Throwing up to start some discussion...

### DIFF
--- a/journal/journal.go
+++ b/journal/journal.go
@@ -1,0 +1,178 @@
+/* Provides a low-level Go interface to the systemd journal C API.
+
+All public methods map closely to the sd-journal API functions. See the
+sd-journal.h documentation[1] for information about each function.
+
+[1] http://www.freedesktop.org/software/systemd/man/sd-journal.html
+*/
+package journal
+
+/*
+#cgo pkg-config: libsystemd-journal
+#include <systemd/sd-journal.h>
+#include <stdlib.h>
+#include <syslog.h>
+
+int go_sd_journal_print(int priority, char* s) {
+	int r;
+	r = sd_journal_print(priority, "%s", s);
+	return r;
+}
+*/
+import "C"
+import (
+	"errors"
+	"fmt"
+	"unsafe"
+)
+
+// Journal entry field strings which correspond to:
+// http://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html
+const (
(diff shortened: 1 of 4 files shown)


journal/journal.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+func (j *Journal) GetData(field string) (string, error) {
+	f := C.CString(field)
+	defer C.free(unsafe.Pointer(f))
+
+	var d unsafe.Pointer
+	var l C.size_t
+
+	err := C.sd_journal_get_data(j.cjournal, f, &d, &l)
```
vcaputo: From the sd_journal_get_data() documentation:
> The returned data is in a read-only memory map and is only valid until the next invocation of sd_journal_get_data() or sd_journal_enumerate_data(), or the read pointer is altered.

Doesn't this imply that we should have a mutex in Journal to serialize access to j.cjournal?

journal/read.go
```diff
@@ -0,0 +1,191 @@
...
+	// TODO: the timeout logic here isn't very good.
+	pollDone := make(chan bool, 1)
+	go func() {
+	poll:
+		for {
+			select {
+			case <-pollDone:
+				break poll
+			default:
+				events <- r.journal.Wait(uint64(time.Duration(2) * time.Second))
```
vcaputo: This doesn't seem safe to me, unless I'm missing something we can't safely enter the sd_journal C api from multiple goroutines on the same journal instance.  This is basically a continuation of my concern mentioned above about a lack of mutex serializing concurrent access to r.journal.cjournal.

journal/journal.go
philips: ack, this seems correct.

vcaputo: I put this together to address the misuse of the C api:
https://github.com/vcaputo/go-systemd/commit/a6aa1c6141da049a42704a7eabf92b502032fd22

/cc @jonboulle @philips

jonboulle: @vcaputo I am swamped with other things, do you want to just put that up as a PR and we can close mine?

vcaputo: Sure, we can do that.

vcaputo: https://github.com/coreos/go-systemd/pull/110

still in rfc form of course

jonboulle:
#110

PR #114, opened 2015-10-13
unit: fix line length check to deal with values

jonboulle:
If an option value is split across multiple lines, the current
line-length check does not catch it as it only considers lines from the
start of them. Hence, we add an additional check while parsing an
option value.

This is not necessary in section headers or option names because they are
alway checked in the core next() loop; it affects values because the
lexer recurses to parse them.

I (and everyone else) missed this while implementing this check
originally because the lexing code is way too complex.


ecnahc515:
Nice! More tests, and more edge cases, all passing. :+1:

vcaputo:
There isn't a recursion issue here, it's just that the lexOptionValueFunc() consumes multiple lines of input rather than revisiting the outer next step loop thingy where the length check is being enforced.

Instead of this band-aid change, why not just change lexOptionValueFunc() to receive the `partial bytes.Buffer` as an argument, then when lexOptionValueFunc() encounters a continuation, it returns itself as the next step with the new accumulation in partial?

Then we would revisit the existing length check on every line, and lexOptionValueFunc() stops being such an oddball.

vcaputo:
@jonboulle Something like this: https://github.com/vcaputo/go-systemd/commit/dff9a78d8a4051c964a69f205d2f9e0183a9ac9f

jonboulle:
That's exactly what I wanted but I was too dumb to realise how to do it. Thanks

jonboulle:
@vcaputo updated

vcaputo:
Likely just a matter of having the luxury of time/bandwidth to stare at it a bit.

LGTM

Surprised my changes pass your tests, I didn't test that shit.

Issue #116, opened 2015-10-14
pkg-config fails to find libsystemd for some docker base-images


jzelinskie:
When attempting to compile an application that depends on go-systemd, I get error messages from pkg-config failing to find `lsystemd`. I've produced this on images derived from `golang:1.5` and `ubuntu:14.04` with dockerfiles along these lines:

```dockerfile
FROM golang:1.5
RUN apt-get update
RUN apt-get -y install build-essentials pkg-config libsystemd-journal-dev
# compile go application here
```

I've gotten a range of different errors by playing with the combination of the many transitional libsystemd packages that I install:

```
# pkg-config --cflags libsystemd
Package libsystemd was not found in the pkg-config search path.
Perhaps you should add the directory containing `libsystemd.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libsystemd' found
pkg-config: exit status 1
```
```
# pkg-config --cflags libsystemd-journal
Package libsystemd-journal was not found in the pkg-config search path.
Perhaps you should add the directory containing `libsystemd-journal.pc'
to the PKG_CONFIG_PATH environment variable
No package 'libsystemd-journal' found
pkg-config: exit status 1
```

```
# github.com/coreos-inc/quay-builder
/usr/local/go/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/usr/bin/ld: cannot find -lsystemd
collect2: error: ld returned 1 exit status
```

jzelinskie:
This should remain open, but now it only effects the `sdjournal` package

jonboulle:
@jzelinskie can you reproduce with this Dockerfile?
```
FROM golang:1.5
RUN apt-get update
RUN apt-get -y install pkg-config libsystemd-journal-dev
RUN go get github.com/coreos/go-systemd/journal github.com/coreos/go-systemd/sdjournal
RUN wget -O main.go https://gist.githubusercontent.com/jonboulle/86132305a24793423ebd/raw/aaceea155212126a4a63f6747c3acb308e8753a4/main.go
RUN go build -a -o journaltester .
```

jzelinskie:
@jonboulle that Dockerfile actually works fine now. I spent a little time trying to repro with another Dockerfile and failed, but I can still repro it with our application. I'll spend some more time on it when I can so that we can figure out this mystery.

PR #110, opened 2015-09-17
[RFC] add journal bindings around sd-journal

vcaputo:
Work in progress. Based on Dan Mace's initial prototype announced here:
http://lists.freedesktop.org/archives/systemd-devel/2014-June/019840.html

Throwing up to start some discussion...

PR #117, opened 2015-10-15
journal: split into read and write packages

### LINKED ISSUE #115: Cgo / pkg-config dependency for Journal Reading
Since https://github.com/coreos/go-systemd/pull/110 go-systemd now uses Cgo, and depends on both `libsystemd-journal` (at build and runtime) and `pkg-config` (at build).  These new dependencies are only used by the code for reading from the systemd journal.

This is moderately inconvenient, as previously go-systemd only used native Go packages to communicate with the journal or systemd.

Ideally native Go could be used, but maybe hiding the journal-reading feature behind a build flag hides the pain.

CC @philips @jonboulle

jonboulle:
0020698a980adf8a2c440d9be709edc85b54544a (via #110) introduced bindings
to read from the journal, which wrap around the sd-journal C API (the
recommended approach to access the journal [1]). Unfortunately that
means that the github.com/coreos/go-systemd/journal package suddenly
moved from being pure Go to having a hard dependency on cgo and
sd-journal, which caused issues on some platforms.

After thinking about this a bit more I think it makes most sense to
split apart the two (read and write) APIs into different packages (which
Zbigniew actually originally suggested a long time ago [3]), as the
implementations are completely separate. This also means we don't need
to futz about with build tag magic to allow people to e.g. only use the
pure-Go part of the API.

Fixes #115 and #116.

[1]: http://lists.freedesktop.org/archives/systemd-devel/2014-June/019904.html
[2]: https://github.com/coreos/go-systemd/issues/116
[3]: http://lists.freedesktop.org/archives/systemd-devel/2014-June/020067.html


journal/journal.go
```diff
@@ -1,4 +1,3 @@
-// Copyright 2015 RedHat, Inc.
```
jonboulle: Sorry, the diff is pretty broken here. These are really just file moves:
`journal/read.go` -> `journal/journal.go`
`journal/journal.go` -> `sdjournal/journal.go`

vcaputo:
seems reasonable, lgtm

PR #118, opened 2015-10-15
sdjournal: import coreos.com/go-systemd/journal in test

vcaputo:
Splitting the C bindings into sdjournal broke the test because
the test uses the native journal api to generate log activity.


jonboulle:
Thanks!

PR #119, opened 2015-10-17
Add interface for Inhibit method

mikkeloscar:
Adds interface for Inhibit method defined here: http://www.freedesktop.org/wiki/Software/systemd/inhibit/

The inhibit method is only useful if it is possible to subscribe to signals on the logind dbus. That's what the `Subscribe` method is for. It can be used like this:

```go
signal := login1Conn.Subscribe("PrepareForSleep")

sig := <-signal
switch sig.Name {
case "org.freedesktop.login1.Manager.PrepareForSleep":
    ....
}

```

Let me know if you think `Subcribe` should be done differently.

### DIFF
--- a/login1/dbus.go
+++ b/login1/dbus.go
@@ -16,6 +16,7 @@
 package login1
 
 import (
+	"fmt"
 	"os"
 	"strconv"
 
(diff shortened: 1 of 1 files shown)


jonboulle:
LGTM from eyeballing, but any chance of a test?

mikkeloscar:
Hmm, how do you mock a dbus interface in order to test it? I can't seem to find any other tests that does something similar.

jonboulle:
You're right, I was thinking about the wrong thing. I think we can land the code as is, but can you please rewrite your commit messages to follow our [format guidelines](https://github.com/coreos/go-systemd/blob/master/CONTRIBUTING.md#format-of-the-commit-message)? (e.g. "login1: add interface for Inhibit method", "login1: add subscribe method"). Thanks!

mikkeloscar:
That should do it, thanks!

login1/dbus_test.go
```diff
@@ -26,3 +26,16 @@ func TestNew(t *testing.T) {
...
 }
+
+// TestInhibit ensures that Inhibit works without errors.
+func TestInhibit(t *testing.T) {
+	conn, err := New()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, err = conn.Inhibit("sleep", "Test program", "Testing inhibitor method", "delay")
```
jonboulle: Er, sorry, can you revert this entirely? An unwanted side effect if people are testing locally!

mikkeloscar: Sorry, I accidentally squashed the commit instead of deleting it, should be fixed now.

jonboulle:
Thanks for the patch!

mikkeloscar:
Thanks for merging, I'm sorry for the confusion.

PR #120, opened 2015-10-17
Use travis container-based infrastructure

mikkeloscar:
Use travis container-based infrastructure for faster builds: http://docs.travis-ci.com/user/workers/container-based-infrastructure/


mikkeloscar:
Nvm, this is obviously not possible since the tests requires root.

PR #121, opened 2015-10-21
Added possibility to use TLS with systemd socket activation

kayrus:
/cc @jonboulle @vcaputo 

Relates to https://github.com/docker/docker/pull/17211

### DIFF
--- a/activation/listeners.go
+++ b/activation/listeners.go
@@ -15,6 +15,7 @@
 package activation
 
 import (
+	"crypto/tls"
 	"net"
 )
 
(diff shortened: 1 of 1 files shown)


activation/listeners.go
```diff
@@ -24,12 +25,16 @@ import (
 // The order of the file descriptors is preserved in the returned slice.
 // Nil values are used to fill any gaps. For example if systemd were to return file descriptors
 // corresponding with "udp, tcp, tcp", then the slice would contain {nil, net.Listener, net.Listener}
```
vcaputo: add something about tlsConfig in the docstring?

vcaputo:
Maybe we should add a new public function for the tls-enabled Listener rather than changing the signature of the existing Listener function.  Changing the signature affects existing callers, and this is a public function...

Maybe change the Listener() you've created to a private listener(), as-is, then wrap it with a public Listener() having the original signature, always passing nil for the tls.Config *, and a new public TLSListener(), passing down a supplied tls.Config *.

jonboulle:
+1 to vc's suggestions

activation/listeners.go
```diff
@@ -35,3 +36,20 @@ func Listeners(unsetEnv bool) ([]net.Listener, error) {
 	}
 	return listeners, nil
 }
+
+// Create TLS-ready TCP listeners
+func ListenersTls(unsetEnv bool, tlsConfig *tls.Config) ([]net.Listener, error) {
```
jonboulle: TLSListeners()

Needs a proper docstring

activation/listeners.go
```diff
@@ -35,3 +36,20 @@ func Listeners(unsetEnv bool) ([]net.Listener, error) {
...
+	listeners, err := Listeners(unsetEnv)
```
jonboulle: Pedantic, but I would rather explicitly return `nil, err` here if Listeners() returns nil, rather than relying on it implicitly returning nil to us

activation/listeners.go
jonboulle: describing that it will only do it for TCP sockets

kayrus:
@jonboulle updated

jonboulle:
LGTM, thanks

PR #123, opened 2015-11-14
Implement AddDisjunction, AddConjunction and FlushMatches

audebert:
See http://www.freedesktop.org/software/systemd/man/sd_journal_add_match.html for more information.

### DIFF
--- a/sdjournal/journal.go
+++ b/sdjournal/journal.go
@@ -107,6 +107,32 @@ func (j *Journal) AddMatch(match string) error {
 	return nil
 }
 
+// AddDisjunction inserts a logical OR in the match list.
+func (j *Journal) AddDisjunction() error {
+	j.mu.Lock()
+	r := C.sd_journal_add_disjunction(j.cjournal)
+	j.mu.Unlock()
+
+	if r < 0 {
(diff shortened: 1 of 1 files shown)


sdjournal/journal.go
```diff
@@ -107,6 +107,39 @@ func (j *Journal) AddMatch(match string) error {
 	return nil
 }
 
+// AddDisjunction inserts a logical OR in the match list.
+func (j *Journal) AddDisjunction() (int, error) {
```
jonboulle: The return value for this and AddConjunction should just be `error` - it's redundant to have `int` as well when it's only used to propagate the return code. (I assume you were emulating the signature of `Next`, but I would argue that its current behaviour in the case of failure -- of returning both a non-zero int and a non-nil error -- is unidiomatic, since Go doesn't typically overload errors into return values the same way as in C).

Unfortunately, that brings up the fact that `AddMatch` isn't properly checking return codes right now and only ever returns `nil`. Would you mind fixing that too? Thanks!

audebert:
Thank you for your comments, I have updated the PR.

jonboulle:
Thanks! It'd be nice to make all the errors more structured at some point, but this lgtm for now.

PR #124, opened 2015-11-27
sdjournal: expose sd_journal_set_data_threshold

runcom:
Signed-off-by: Antonio Murdaca <runcom@redhat.com>

### DIFF
--- a/sdjournal/journal.go
+++ b/sdjournal/journal.go
@@ -220,6 +220,22 @@ func (j *Journal) GetData(field string) (string, error) {
 	return msg, nil
 }
 
+// SetDataThresold sets the data field size threshold for data returned by
+// GetData. To retrieve the complete data fields this threshold should be
+// turned off by setting it to 0, so that the library always returns the
+// complete data objects.
+func (j *Journal) SetDataThreshold(threshold uint64) error {
+	j.mu.Lock()
+	r := C.sd_journal_set_data_threshold(j.cjournal, C.size_t(threshold))
+	j.mu.Unlock()
+
+	if r < 0 {
+		return fmt.Errorf("failed to set data threshold: %d", r)
+	}
+
+	return nil
+}
+
 // GetRealtimeUsec gets the realtime (wallclock) timestamp of the current
 // journal entry.
 func (j *Journal) GetRealtimeUsec() (uint64, error) {
--- a/sdjournal/read.go
+++ b/sdjournal/read.go
@@ -37,8 +37,6 @@ type JournalReaderConfig struct {
 	// Show only journal entries whose fields match the supplied values. If
 	// the array is empty, entries will not be filtered.
 	Matches []Match
-
-	MaxEntryLength uint64 // the max allowable length for an entry (default is 64K)
 }
 
 // JournalReader is an io.ReadCloser which provides a simple interface for iterating through the
@@ -50,9 +48,9 @@ type JournalReader struct {
 // NewJournalReader creates a new JournalReader with configuration options that are similar to the
 // systemd journalctl tool's iteration and filtering features.
 func NewJournalReader(config JournalReaderConfig) (*JournalReader, error) {
-	var err error
 	r := &JournalReader{}
 
+	var err error
 	// Open the journal
 	if r.journal, err = NewJournal(); err != nil {
 		return nil, err


sdjournal/read.go
```diff
@@ -38,7 +38,10 @@ type JournalReaderConfig struct {
 	// the array is empty, entries will not be filtered.
 	Matches []Match
 
-	MaxEntryLength uint64 // the max allowable length for an entry (default is 64K)
+	// explicitly set MaxEntryLenght to 0, which means unlimited
+	DisableMaxEntryLength bool
+	// the max allowable length for an entry (default is 64K)
+	MaxEntryLength uint64
```
runcom: disabling this was tricky because unless this becomes a pointer there's no way of saying if the used want to disable it because it will default to 0 when the struct is initialized. I thought about creating a `NewJournalReaderConfig` but was kind of ugly.

sdjournal/journal.go
```diff
@@ -220,6 +220,18 @@ func (j *Journal) GetData(field string) (string, error) {
 	return msg, nil
 }
 
+// SetDataThresold sets the data field size threshold for data returned by
+// GetData. To retrieve the complete data fields this threshold should be
+// turned off by setting it to 0, so that the library always returns the
+// complete data objects.
+func (j *Journal) SetDataThreshold(threshold uint64) error {
+	r := C.sd_journal_set_data_threshold(j.cjournal, C.size_t(threshold))
```
vcaputo: serialize the use of j.cjournal like the rest of the methods

sdjournal/read.go
jonboulle: Since this value was hitherto not actually used, you don't have to worry about it being backwards-compatible, so let's just give it the same behaviour as the systemd threshold: 0 means unlimited.

runcom: Perfect thanks I'll update it

sdjournal/journal.go
runcom: Will do, thanks overlooked

sdjournal/read.go
runcom: EDIT: sorry, but how do you mimic the systemd function if someone initialize this Config it will default to unlimited, not 64k as the systemd lib

runcom: Unless a pointer is used to say the value is empty so use a default. But with a pointer struct initialization will be awful for a numeric type

jonboulle: Ugh, you're right. Back to the drawing board, let me ponder it a little.

runcom: @jonboulle any news? isn't the way I went good at all?

jonboulle: @runcom How about just scrapping this bit of the config struct entirely, and requiring users to call `SetDataThreshold()` explicitly after they instantiate a `JournalReader`? That seems like the cleanest path forward, imo.

runcom: Agree, I'll update the patch

runcom:
@vcaputo @jonboulle updated, PTAL

sdjournal/journal.go
```diff
@@ -220,6 +220,22 @@ func (j *Journal) GetData(field string) (string, error) {
...
+// SetDataThresold sets the data field size threshold for data returned by
+// GetData. To retrieve the complete data fields this threshold should be
+// turned off by setting it to 0, so that the library always returns the
+// complete data objects.
+func (j *Journal) SetDataThreshold(threshold uint64) error {
+	j.mu.Lock()
+	r := C.sd_journal_set_data_threshold(j.cjournal, C.size_t(threshold))
+	j.mu.Unlock()
+
+	if r != 0 {
```
vcaputo: Everywhere else we're testing the return value from sd-journal using `if r < 0 {`, I see no reason to deviate here.

jonboulle: Not quite true, we're a bit inconsistent: https://github.com/coreos/go-systemd/pull/124/files#diff-167dc1dd14a090cd485c7dfbb2cfa9b3R262 , https://github.com/coreos/go-systemd/pull/124/files#diff-167dc1dd14a090cd485c7dfbb2cfa9b3R276

jonboulle: But I agree we should use r < 0 here

jonboulle:
LGTM after vito's comment is addressed

sdjournal/journal.go
vcaputo: *sigh*  I didn't check all of them, we should make them consistent.

runcom:
@vcaputo updated

vcaputo:
Thanks @runcom, lgtm!

PR #126, opened 2015-11-28
Add method to obtain systemd1.manager interface properties

ranjib:
This will allow go-systemd users to obtain metadata information like version. etc
```go
bus, _ := dbus.New()
systemdVersion, _ := bus.GetManagerProperty("Version")
```

### DIFF
--- a/dbus/dbus.go
+++ b/dbus/dbus.go
@@ -148,6 +148,17 @@ func newConnection(createBus func() (*dbus.Conn, error)) (*Conn, error) {
 	return c, nil
 }
 
+// GetManagerProperty returns the value of a property on the org.freedesktop.systemd1.Manager
+// interface. The value is returned in its string representation, as defined at
+// https://developer.gnome.org/glib/unstable/gvariant-text.html
+func (c *Conn) GetManagerProperty(prop string) (string, error) {
+	variant, err := c.sysobj.GetProperty("org.freedesktop.systemd1.Manager." + prop)
+	if err != nil {
+		return "", err
+	}
+	return variant.String(), nil
+}
+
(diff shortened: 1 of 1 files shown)


dbus/dbus.go
```diff
@@ -148,6 +148,15 @@ func newConnection(createBus func() (*dbus.Conn, error)) (*Conn, error) {
 	return c, nil
 }
 
+// GetProperty returns values of a org.freedesktop.systemd1.Manager interface property
+func (c *Conn) GetProperty(prop string) (string, error) {
```
jonboulle: Since the `Conn` is pretty general, I think it might be a little clearer to call this `GetManagerProperty` (c.f. the existing `GetUnitProperty` for example) - does that seem reasonable? Other than that, LGTM

ranjib: what about GetManagerProperty  ? my understanding is that we can use it to fetch only systemd manager specific property, like Version, Features, Logtarget etc. Also, some of these returns non-string (array, int etc). Should i instead use interface{} as return type or something more appropriate ?

jonboulle: Oh yes, good call. How about something like:

```
// GetManagerProperty returns the value of a property on the org.freedesktop.systemd.Manager 
// interface. The value is returned in its string representation, as defined at 
// https://developer.gnome.org/glib/unstable/gvariant-text.html
func (c *Conn) GetManagerProperty(prop string) (string, error) 
```

jonboulle:
Ping?

ranjib:
@jonboulle done

jonboulle:
thanks!

PR #127, opened 2015-12-17
util: add ErrSoNotFound error var

iaguis:
This allows rkt to detect that the problem was that we couldn't open
libsystemd and just warn the user.

For example, NixOS has shared libraries in weird paths and we don't want
to just fail if we can't find it. See
https://github.com/coreos/rkt/issues/1879

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -65,6 +65,8 @@ import (
 	"unsafe"
 )
 
+var ErrSoNotFound = errors.New("unable to open a handle to libsystemd")
+
 // libHandle represents an open handle to the systemd C library
 type libHandle struct {
 	handle  unsafe.Pointer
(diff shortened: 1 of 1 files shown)


alban:
LGTM

jonboulle:
Why a type and not just a declared error var?

iaguis:
> Why a type and not just a declared error var?

A type is what came to mind but I don't mind (:D) changing it to an error var.

jonboulle:
Please, it's more idiomatic (eg stdlib)

Iago López Galeiras <notifications@github.com> schrieb am Do., 17. Dez.
2015 16:18:

iaguis:
OK. Updated

PR #128, opened 2015-12-19
Enable reading of the journal from a custom directory

chrissnell:
This PR enables the user to read the system journal from a custom (alternate) directory by offering an interface to the ```sd_journal_open_directory()``` function.

It also adds a ```GetMachineID()``` convenience function, useful for typical journald storage directories that include the machine-id in the path.

### DIFF
--- a/sdjournal/journal.go
+++ b/sdjournal/journal.go
@@ -25,7 +25,7 @@
 package sdjournal
 
 /*
-#cgo pkg-config: libsystemd-journal
+#cgo pkg-config: libsystemd
 #include <systemd/sd-journal.h>
 #include <stdlib.h>
 #include <syslog.h>
@@ -81,6 +81,26 @@ func NewJournal() (*Journal, error) {
 
 	if err < 0 {
 		return nil, fmt.Errorf("failed to open journal: %s", err)
+	}
+
+	return j, nil
+}
+
+// NewJournalFromDir returns a new Journal instance pointing to a journal residing
+// in a given directory.  Typically, the directory will be appended with /<machine-id>
+// which can be generated from util.GetMachineID() from this same library.
+func NewJournalFromDir(path string) (*Journal, error) {
+	if len(path) == 0 {
+		return nil, fmt.Errorf("failed to open journal; provided path is nil:")
+	}
+
+	p := C.CString(path)
+	defer C.free(unsafe.Pointer(p))
+
+	j := &Journal{}
+	errInt := C.sd_journal_open_directory(&j.cjournal, p, 0)
+	if errInt < 0 {
+		return nil, fmt.Errorf("failed to open journal in directory %v: %v", path, errInt)
 	}
 
 	return j, nil
--- a/sdjournal/journal.go
+++ b/sdjournal/journal.go
@@ -33,6 +33,7 @@ package sdjournal
 import "C"
 import (
 	"fmt"
+	"path/filepath"
 	"sync"
 	"time"
 	"unsafe"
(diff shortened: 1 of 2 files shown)


chrissnell:
My system runs a newer version of systemd which apparently has merged the libsystemd-journal library into the main libsystemd library.  This PR adjusts the cgo pkg-config accordingly.  Users of go-systemd on newer machines are unlikely to be able to build this library without this change.

util/util.go
```diff
@@ -254,3 +255,19 @@ func IsRunningSystemd() bool {
 	}
 	return fi.IsDir()
 }
+
+// GetMachineID returns a host's 128-bit machine ID as a string.
+func GetMachineID() (machineID string, err error) {
```
jonboulle: A few style nits for this function:
a) I'd omit the named result parameters - they add little value and it stutters a bit in godoc
b) Error strings should start [lower-case](https://github.com/golang/go/wiki/CodeReviewComments#error-strings)
b) I would rather promote the comment to the docstring, similar to that of `IsRunningSystemd`, e.g.: "This functions similar to systemd's `sd_id28_get_machine`, ..." 
c) Let's just `strings.TrimSpace` instead of manual string slicing

util/util.go
```diff
@@ -254,3 +255,19 @@ func IsRunningSystemd() bool {
...
+	// There is a systemd library function, sd_id128_get_machine, which can be used
+	// to obtain the machine ID.  However, that function simply reads the string
+	// from /etc/machine-id on the filesystem. To simplify things, we'll just read
+	// it directly from the file, too.
+	systemID, err := ioutil.ReadFile("/etc/machine-id")
+	if err != nil {
+		fmt.Errorf("Failed to read /etc/machine-id: %v", err)
```
jonboulle: This isn't actually doing something, `fmt.Errorf` needs to be assigned (but per my comment about named result parameters let's just return)

sdjournal/journal.go
```diff
@@ -86,6 +86,26 @@ func NewJournal() (*Journal, error) {
 	return j, nil
 }
 
+// NewJournalFromDir returns a new Journal instance pointing to a journal residing
+// in a given directory.  Typically, the directory will be appended with /<machine-id>
+// which can be generated from util.GetMachineID() from this same library.
+func NewJournalFromDir(path string) (*Journal, error) {
+	if len(path) == 0 {
+		return nil, fmt.Errorf("failed to open journal; provided path is nil:")
```
jonboulle: Any reason to do this and not just let `sd_journal_open_directory` error out itself? Otherwise, we're required to update if they happen to change the API there and e.g. support empty string indicating current directory, or something like that.
Coming at it from the opposite direction, this could be an `AbsPath` check instead.

jonboulle:
> My system runs a newer version of systemd which apparently has merged the libsystemd-journal library into the main libsystemd library. This PR adjusts the cgo pkg-config accordingly. Users of go-systemd on newer machines are unlikely to be able to build this library without this change.

that's kinda annoying :-(. What system are you on out of interest?

jonboulle:
thanks! Some style nits, but :+1: overall

chrissnell:
@jonboulle - fixed up those issues for you.  Does it look okay now?  :smiley:

chrissnell:
@jonboulle Regarding my build platform, I'm building on Ubuntu 15.10.  For what it's worth, it's the same situation on my CoreOS 877.1.0 servers:

```
core@coreos-005 ~ $ ls -l /lib/libsystem*
lrwxrwxrwx 1 root root     20 Nov 26 03:18 /lib/libsystemd.so -> libsystemd.so.0.10.2
lrwxrwxrwx 1 root root     20 Nov 26 03:18 /lib/libsystemd.so.0 -> libsystemd.so.0.10.2
-rwxr-xr-x 1 root root 552696 Nov 26 02:36 /lib/libsystemd.so.0.10.2
```

There's no libsystemd-journal to link against.

jonboulle:
Thanks, I have a couple more style nits but I'm just going to chase them up myself.

Re: pkg-config, they both resolve fine on my Fedora 22 system, I think this is okay to roll forward with:
```
: trevize ~ ; pkg-config --libs libsystemd
-lsystemd 
: trevize ~ ; pkg-config --libs libsystemd-journal
-lsystemd 
```

jonboulle:
Landed via #130

PR #129, opened 2015-12-19
Add sdjournal.WaitIndefinitely() method

chrissnell:
Add a method to wait indefinitely for a new journal entry.   This is useful for writing programs that "tail" the journal and should not timeout waiting for a new entry.

### DIFF
--- a/sdjournal/journal.go
+++ b/sdjournal/journal.go
@@ -57,6 +57,11 @@ const (
 	SD_JOURNAL_INVALIDATE = int(C.SD_JOURNAL_INVALIDATE)
 )
 
+// The maximum value for a time.Duration.  Used for sdjournal.WaitIndefinitely()
+const (
+	maxDuration = "2540400h10m10.000000000s"
+)
+
 // A Journal is a Go wrapper of an sd_journal structure.
 type Journal struct {
 	cjournal *C.sd_journal
@@ -294,14 +299,8 @@ func (j *Journal) Wait(timeout time.Duration) int {
 // WaitIndefinitely will synchronously wait until the journal gets changed.  This
 // call will sleep indefinitely while it waits for a change.
 func (j *Journal) WaitIndefinitely() int {
-        j.mu.Lock()
-        // sd_journal_wait(3) calls for a (uint64_t) -1 to be passed to signify
-        // indefinite wait, but using a -1 overflows our C.uint64_t, so we use an
-        // equivalent hex value
-        r := C.sd_journal_wait(j.cjournal, (C.uint64_t)(0xffffffffffffffff))
-        j.mu.Unlock()
-
-        return int(r)
+	indef, _ := time.ParseDuration(maxDuration)
+	return j.Wait(indef)
 }
 
 // GetUsage returns the journal disk space usage, in bytes.
--- a/sdjournal/journal.go
+++ b/sdjournal/journal.go
@@ -291,6 +291,19 @@ func (j *Journal) Wait(timeout time.Duration) int {
 	return int(r)
 }
 
+// WaitIndefinitely will synchronously wait until the journal gets changed.  This
+// call will sleep indefinitely while it waits for a change.
+func (j *Journal) WaitIndefinitely() int {
+        j.mu.Lock()
+        // sd_journal_wait(3) calls for a (uint64_t) -1 to be passed to signify
+        // indefinite wait, but using a -1 overflows our C.uint64_t, so we use an
+        // equivalent hex value
+        r := C.sd_journal_wait(j.cjournal, (C.uint64_t)(0xffffffffffffffff))
+        j.mu.Unlock()
+
+        return int(r)
+}
+
 // GetUsage returns the journal disk space usage, in bytes.
 func (j *Journal) GetUsage() (uint64, error) {
 	var out C.uint64_t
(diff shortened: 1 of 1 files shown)


sdjournal/journal.go
```diff
@@ -291,6 +291,19 @@ func (j *Journal) Wait(timeout time.Duration) int {
...
 }
 
+// WaitIndefinitely will synchronously wait until the journal gets changed.  This
+// call will sleep indefinitely while it waits for a change.
+func (j *Journal) WaitIndefinitely() int {
+        j.mu.Lock()
+        // sd_journal_wait(3) calls for a (uint64_t) -1 to be passed to signify
+        // indefinite wait, but using a -1 overflows our C.uint64_t, so we use an
+        // equivalent hex value
+        r := C.sd_journal_wait(j.cjournal, (C.uint64_t)(0xffffffffffffffff))
```
jonboulle: Alternatively we could just do this in CGo:
```
/*
int my_journal_wait_indefinitely(sd_journal* j) {
        return sd_journal_wait(j, (uint64_t) -1);
}
*/
import "C"

...

r := C.my_journal_wait_indefinitely(j.cjournal)
```

Not sure if there's a reason to prefer one way or another - @vcaputo ?

chrissnell:
See: https://busybox.net/~landley/c99-draft.html#6.3.1.3

> [...] if the new type is unsigned, the value is converted by repeatedly adding or subtracting one more than the maximum value that can be represented in the new type until the value is in the range of the new type.

I believe that using ```0xffffffffffffffff``` is dependable and valid but I think your method is more legible, especially if we add a comment that explains why we're doing this.  LMK which you prefer and I will make the change if necessary.

sdjournal/journal.go
vcaputo: @jonboulle How about we just drop this function altogether and instead define an infinite duration value for Wait() which then supplies the appropriate hex value?

jonboulle: @vcaputo I thought of that but what would an appropriate value be?

vcaputo: @jonboulle 290 years?

chrissnell: https://golang.org/src/time/time.go?s=1855:2369#L463

Also:

https://golang.org/src/time/time.go?s=1855:2369#L434

vcaputo: There's no practically valid scenario where you'd actually wait for journal logs for hundreds of years as opposed to indefinitely, so I think it's perfectly safe to turn the maximum value time.Duration can represent into infinity.

jonboulle: SGTM, who wants to implement that? ;-)

chrissnell: I'm not following how you want this done.  Wait() takes a time; do you want a WaitIndefinitely() helper function that passes 2540400h10m10.000000000s to Wait()?

jonboulle: I'll let @vcaputo speak for himself but that's what I had in mind, except
with some sugar, e.g. could expose an `IndefiniteWait` sentinel const...

On Wed, Dec 23, 2015 at 6:44 PM, Christopher Snell <notifications@github.com

sdjournal/journal.go
```diff
@@ -57,6 +57,11 @@ const (
 	SD_JOURNAL_INVALIDATE = int(C.SD_JOURNAL_INVALIDATE)
 )
 
+// The maximum value for a time.Duration.  Used for sdjournal.WaitIndefinitely()
+const (
+	maxDuration = "2540400h10m10.000000000s"
```
vcaputo: Make this public, it's part of the Wait() api.  I'd rather it be like what @jonboulle suggested, IndefiniteWait, or InfiniteDuration... something like that.

sdjournal/journal.go
```diff
@@ -291,6 +296,13 @@ func (j *Journal) Wait(timeout time.Duration) int {
 	return int(r)
 }
 
+// WaitIndefinitely will synchronously wait until the journal gets changed.  This
+// call will sleep indefinitely while it waits for a change.
+func (j *Journal) WaitIndefinitely() int {
+	indef, _ := time.ParseDuration(maxDuration)
+	return j.Wait(indef)
+}
```
vcaputo: The Wait() implementation needs to change to detect when the sentinel value was supplied and supply the 0xffff... value to sd_journal_wait() when it does.  I didn't mean for us to just supply a large value to Wait() and treat that as unlimited, there is likely an implementation difference underneath - one registers a timer, the other doesn't.  Let's not involve a timer when we're interested in waiting forever.

Please document the sentinel in the Wait() comment as well.

As for WaitIndefinitely(), I don't think we need it, just pass the public sentinel value to Wait() for the indefinite wait.

chrissnell:
OK, guys.  Hope this works better for you, @vcaputo @jonboulle

vcaputo:
works for me, @jonboulle ?

jonboulle:
Mostly LGTM but a few style issues - to save the back-and-forth, I've chased them up in #134 and will land there. (I also changed the [exact implementation of IndefiniteWait](https://github.com/coreos/go-systemd/pull/134/files#diff-167dc1dd14a090cd485c7dfbb2cfa9b3R65) as I think it's best to follow upstream). For future reference, please try to follow the commit message style and godoc style wherever possible. Thanks for the patch!

PR #131, opened 2015-12-21
un-break errInt->err remplacement

lyonel:


jonboulle:
*sigh* thanks for catching, clearly need more tests. I took this a bit further via #132, will land there.

PR #136, opened 2016-01-05
Add journal.GetDataMap to retreive the full key-value map of an entry.

wrouesnel:
Implements the same semantics as the JOURNAL_FOREACH_DATA_RETVAL macro in journal-internal.h and produces a map[string]string of the full contents of every field.

This matches the output of `journalctl -o export` as strings, but allows manipulation of the value as a map (and exploring the full range of fields available).

### DIFF
--- a/sdjournal/journal.go
+++ b/sdjournal/journal.go
@@ -44,13 +44,16 @@ import (
 // Journal entry field strings which correspond to:
 // http://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html
 const (
-	SD_JOURNAL_FIELD_SYSTEMD_UNIT = "_SYSTEMD_UNIT"
-	SD_JOURNAL_FIELD_MESSAGE      = "MESSAGE"
-	SD_JOURNAL_FIELD_PID          = "_PID"
-	SD_JOURNAL_FIELD_UID          = "_UID"
-	SD_JOURNAL_FIELD_GID          = "_GID"
-	SD_JOURNAL_FIELD_HOSTNAME     = "_HOSTNAME"
-	SD_JOURNAL_FIELD_MACHINE_ID   = "_MACHINE_ID"
+	SD_JOURNAL_FIELD_SYSTEMD_UNIT        = "_SYSTEMD_UNIT"
+	SD_JOURNAL_FIELD_MESSAGE             = "MESSAGE"
+	SD_JOURNAL_FIELD_PID                 = "_PID"
+	SD_JOURNAL_FIELD_UID                 = "_UID"
+	SD_JOURNAL_FIELD_GID                 = "_GID"
+	SD_JOURNAL_FIELD_HOSTNAME            = "_HOSTNAME"
+	SD_JOURNAL_FIELD_MACHINE_ID          = "_MACHINE_ID"
+	SD_JOURNAL_FIELD_REALTIME_TIMESTAMP  = "__REALTIME_TIMESTAMP"
+	SD_JOURNAL_FIELD_MONOTONIC_TIMESTAMP = "__MONOTONIC_TIMESTAMP"
+	SD_JOURNAL_FIELD_CURSOR              = "__CURSOR"
 )
 
 // Journal event constants
@@ -262,7 +265,7 @@ func (j *Journal) GetDataValue(field string) (string, error) {
 	return strings.SplitN(val, "=", 2)[1], nil
 }
 
-// GetDataMap returns all the Key-Value pairs of data for the current journal
+// GetJournalEntry returns all the Key-Value pairs of data for the current journal
 // entry. Mimics the behavior of output_export in logs-show.c
 func (j *Journal) GetDataMap() (map[string]string, error) {
 	var r C.int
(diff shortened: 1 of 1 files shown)


sdjournal/journal.go
```diff
@@ -260,6 +262,61 @@ func (j *Journal) GetDataValue(field string) (string, error) {
...
+
+	resultMap := make(map[string]string)
+
+	j.mu.Lock()
+	defer j.mu.Unlock()
+
+	var realtimeUsec C.uint64_t
+	r = C.sd_journal_get_realtime_usec(j.cjournal, &realtimeUsec)
+	if r < 0 {
+		return nil, fmt.Errorf("Failed to get realtime timestamp: %d", r)
```
jonboulle: errors should start lower-case

sdjournal/journal.go
```diff
@@ -260,6 +262,61 @@ func (j *Journal) GetDataValue(field string) (string, error) {
...
+	}
+	resultMap["__REALTIME_TIMESTAMP"] = fmt.Sprintf("%d", realtimeUsec)
+
+	var monotonicUsec C.uint64_t
+	var bootid C.sd_id128_t // We don't do anything with this since we get it below
+	r = C.sd_journal_get_monotonic_usec(j.cjournal, &monotonicUsec, &bootid)
+	if r < 0 {
+		return nil, fmt.Errorf("Failed to get realtime timestamp: %d", r)
```
jonboulle: copypastad error message

sdjournal/journal.go
```diff
@@ -260,6 +262,61 @@ func (j *Journal) GetDataValue(field string) (string, error) {
...
+	}
+	resultMap["__MONOTONIC_TIMESTAMP"] = fmt.Sprintf("%d", monotonicUsec)
+
+	var cursor *C.char
+	r = C.sd_journal_get_cursor(j.cjournal, &cursor)
+	if r < 0 {
+		return nil, fmt.Errorf("Failed to get realtime timestamp: %d", r)
```
jonboulle: copypastad error message

sdjournal/journal.go
```diff
@@ -260,6 +262,61 @@ func (j *Journal) GetDataValue(field string) (string, error) {
...
+	}
+	resultMap["__CURSOR"] = C.GoString(cursor)
+
+	// Implements the JOURNAL_FOREACH_DATA_RETVAL macro from journal-internal.h
+	C.sd_journal_restart_data(j.cjournal)
+	for {
+		r = C.sd_journal_enumerate_data(j.cjournal, &d, &l)
+		if r <= 0 {
```
jonboulle:
```
if r == 0 {
  break
}
if r < 0 {
  ...
}
```

sdjournal/journal.go
```diff
@@ -260,6 +262,61 @@ func (j *Journal) GetDataValue(field string) (string, error) {
...
+
+	j.mu.Lock()
+	defer j.mu.Unlock()
+
+	var realtimeUsec C.uint64_t
+	r = C.sd_journal_get_realtime_usec(j.cjournal, &realtimeUsec)
+	if r < 0 {
+		return nil, fmt.Errorf("Failed to get realtime timestamp: %d", r)
+	}
+	resultMap["__REALTIME_TIMESTAMP"] = fmt.Sprintf("%d", realtimeUsec)
```
jonboulle: can you split these into consts as we have for https://github.com/coreos/go-systemd/pull/136/files#diff-167dc1dd14a090cd485c7dfbb2cfa9b3L42 ?

sdjournal/journal.go
```diff
@@ -260,6 +262,61 @@ func (j *Journal) GetDataValue(field string) (string, error) {
...
+	var r C.int
+	var d unsafe.Pointer
+	var l C.size_t
+
+	resultMap := make(map[string]string)
+
+	j.mu.Lock()
+	defer j.mu.Unlock()
+
+	var realtimeUsec C.uint64_t
```
jonboulle: I am not quite sure how I feel about muxing these fields in with the data map - from http://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html#Address%20Fields : 
>  Note that these are not proper fields when stored in the journal but for addressing metadata of entries. 

perhaps we could have something like
```
type JournalEntry {
   fields map[string]string
   Cursor string
   RealtimeTimestamp uint64
   MonotonicTimestamp uint64
}
```
or similar, then instead of GetDataMap, `GetJournalEntry() (*JournalEntry, error)`

@vcaputo any thoughts?

wrouesnel: The idea was to replicate the behavior of the journalctl export/json format messages. This would be a valid object if json serialized to be dumped into another journal. Although, the structure t implementation would allow keeping this behavior by default (and .String() could implement the export format)

vcaputo: Haven't taken a close enough look at this to have an informed opinion

jonboulle: @vcaputo how about now?

wrouesnel:
Bump: tackled all the cosmetic stuff.

wolfeidau:
Would love it if this was merged, just started using this library and this would enable me to delete some code.

Cheers

sdjournal/journal.go
```diff
@@ -260,6 +265,63 @@ func (j *Journal) GetDataValue(field string) (string, error) {
 	return strings.SplitN(val, "=", 2)[1], nil
 }
 
+// GetJournalEntry returns all the Key-Value pairs of data for the current journal
```
jonboulle: This doesn't match the function name

jonboulle:
personally I'm still not sure about https://github.com/coreos/go-systemd/pull/136/files#r49264989

lucab:
Closed via #164.

PR #137, opened 2016-01-12
sdjournal: Added missing strings package.

jareksm:
sdjournal failed to compile.


jonboulle:
Wow, really? Shame on me. Thanks for the fix.

PR #139, opened 2016-01-13
sdjournal: Reuse buffer to speed up the Follow().

yifan-gu:
Previously, Follow() is pretty slow due to the unnecessary creation of the buffer in each read.


yifan-gu:
cc @jonboulle

jonboulle:
lgtm, thanks

PR #140, opened 2016-01-27
Close the tempfile when journaling large messages

nalind:
When we fail to log a message to journald due to ENOBUFS or EMSGSIZE, we attempt to pass the message to journald by passing it a descriptor attached to some shared memory.  Since we won't be touching it again, we need to close the descriptor after we send it.


jonboulle:
`pkg journal` is pure Go. We could bind to `sd_journal_sendv` in `pkg sdjournal`

jonboulle:
LGTM, thanks.

PR #143, opened 2016-03-03
sdjournal: add support for cursors (get&seek&test)

glerchundi:
This implements/wraps `int sd_journal_seek_cursor(sd_journal* j, const char * cursor);` which allows us to seek by cursor.

### DIFF
--- a/sdjournal/journal.go
+++ b/sdjournal/journal.go
@@ -155,6 +155,24 @@ package sdjournal
 // }
 //
 // int
+// my_sd_journal_get_cursor(void *f, sd_journal *j, char **cursor)
+// {
+//   int (*sd_journal_get_cursor)(sd_journal *, char **);
+//
+//   sd_journal_get_cursor = f;
+//   return sd_journal_get_cursor(j, cursor);
+// }
+//
+// int
+// my_sd_journal_test_cursor(void *f, sd_journal *j, const char *cursor)
+// {
+//   int (*sd_journal_test_cursor)(sd_journal *, const char *);
+//
+//   sd_journal_test_cursor = f;
+//   return sd_journal_test_cursor(j, cursor);
+// }
+//
+// int
 // my_sd_journal_get_realtime_usec(void *f, sd_journal *j, uint64_t *usec)
 // {
 //   int (*sd_journal_get_realtime_usec)(sd_journal *, uint64_t *);
@@ -179,6 +197,16 @@ package sdjournal
 //
 //   sd_journal_seek_tail = f;
 //   return sd_journal_seek_tail(j);
+// }
+//
+//
+// int
+// my_sd_journal_seek_cursor(void *f, sd_journal *j, const char *cursor)
+// {
+//   int (*sd_journal_seek_cursor)(sd_journal *, const char *);
+//
+//   sd_journal_seek_cursor = f;
+//   return sd_journal_seek_cursor(j, cursor);
 // }
 //
 // int
@@ -596,6 +624,50 @@ func (j *Journal) GetRealtimeUsec() (uint64, error) {
 	return uint64(usec), nil
 }
 
+// GetCursor gets the cursor of the current journal entry.
+func (j *Journal) GetCursor() (string, error) {
+	var d *C.char
+
+	sd_journal_get_cursor, err := j.getFunction("sd_journal_get_cursor")
+	if err != nil {
+		return "", err
+	}
+
+	j.mu.Lock()
+	r := C.my_sd_journal_get_cursor(sd_journal_get_cursor, j.cjournal, &d)
+	j.mu.Unlock()
+
+	if r < 0 {
+		return "", fmt.Errorf("failed to get cursor: %d", syscall.Errno(-r))
+	}
+
+	cursor := C.GoString(d)
+
+	return cursor, nil
+}
+
+// TestCursor checks whether the current position in the journal matches the
+// specified cursor
+func (j *Journal) TestCursor(cursor string) error {
+	sd_journal_test_cursor, err := j.getFunction("sd_journal_test_cursor")
+	if err != nil {
+		return err
+	}
+
+	c := C.CString(cursor)
+	defer C.free(unsafe.Pointer(c))
+
+	j.mu.Lock()
+	r := C.my_sd_journal_test_cursor(sd_journal_test_cursor, j.cjournal, c)
+	j.mu.Unlock()
+
+	if r < 0 {
+		return fmt.Errorf("failed to test to cursor %q: %d", cursor, syscall.Errno(-r))
+	}
+
+	return nil
+}
+
 // SeekHead seeks to the beginning of the journal, i.e. the oldest available
 // entry.
 func (j *Journal) SeekHead() error {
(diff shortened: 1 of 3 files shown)


lucab:
Thanks for this PR!
I'm sorry about the wrong timing, but after #141 this needs a bit of reworking to move to `dlopen` style of functions calling.

sdjournal/journal.go
```diff
@@ -293,6 +293,23 @@ func (j *Journal) GetRealtimeUsec() (uint64, error) {
...
+// GetCursor gets the cursor of the current journal entry.
+func (j *Journal) GetCursor() (string, error) {
+	var d *C.char
+
+	j.mu.Lock()
+	r := C.sd_journal_get_cursor(j.cjournal, &d)
+	j.mu.Unlock()
+
+	if r < 0 {
+		return "", fmt.Errorf("failed to get cursor: %d", r)
```
lucab: I think you can wrap an (inverted) `r` into a `syscall.Errno` and expose a more meaningful error to users.

sdjournal/journal.go
```diff
@@ -321,6 +338,22 @@ func (j *Journal) SeekRealtimeUsec(usec uint64) error {
...
+func (j *Journal) SeekCursor(cursor string) error {
+	c := C.CString(cursor)
+	defer C.free(unsafe.Pointer(c))
+
+	j.mu.Lock()
+	r := C.sd_journal_seek_cursor(j.cjournal, c)
+	j.mu.Unlock()
+
+	if r < 0 {
+		return fmt.Errorf("failed to seek to cursor %q: %d", cursor, r)
```
lucab: Same `r` wrapping could be applied here.

glerchundi:
@lucab don't worry! glad to see someone is retaking these pulls.

Tests are passing, rebased, squashed and updated to use dlopened version. Take a look and tell me if you need something else.

Thanks!

sdjournal/journal.go
```diff
@@ -587,6 +606,28 @@ func (j *Journal) GetRealtimeUsec() (uint64, error) {
 	return uint64(usec), nil
 }
 
+// GetCursor gets the cursor of the current journal entry.
+func (j *Journal) GetCursor() (string, error) {
+	var d *C.char
+
+	sd_journal_get_cursor, err := j.getFunction("sd_journal_get_cursor")
+	if err != nil {
+		return 0, err
```
lucab: Return type mismatch, should be `""` instead of `0`.

lucab:
Oh, that was quick! :smiley:
One issue already pointed out inline, and two overall questions:
 * Doesn't your usecase also involve `sd_journal_test_cursor()`? I feel like it is missing here to provide proper cursor feature. Do you think you can add it to this PR?
 * Can you add a test for whole cursor story? (I think having the `test_cursor` will actually make it easier to do get/seek/test and check)

glerchundi:
I'll try throughout this week.

sdjournal/journal.go
glerchundi: ups, fixed. Although I don't know why first go test didn't tell me anything...

glerchundi:
> Doesn't your usecase also involve sd_journal_test_cursor()? I feel like it is missing here to provide proper cursor feature. Do you think you can add it to this PR?

Can be used but I don't, I'm using these methods for a journald forwarder which persists an existing cursor while it's propagating everything else to another centralized logging. So I just need get (current in order to persist) and set (so that it restarts in a specific position, if it fails i fallback to the first journald entry). 

> Can you add a test for whole cursor story? (I think having the test_cursor will actually make it easier to do get/seek/test and check)

seems reasonable, as i said let me check if i can work on it during this week.

lucab:
> Can be used but I don't, I'm using these methods for a journald forwarder which persists an existing cursor while it's propagating everything else to another centralized logging. So I just need get (current in order to persist) and set (so that it restarts in a specific position, if it fails i fallback to the first journald entry).

I see. Skimming through the manpage, I feel like the most common usage would be get/seek/test, but your usecase also makes sense.

> seems reasonable, as i said let me check if i can work on it during this week.

Great! Let us know if you are time-starved for that part, perhaps we can first land this and then complete the whole.

glerchundi:
@lucab i've reworked it a little and now everything seems to work as expected. PTAL to new "TestJournalCursorGetSeekAndTest" test too. I didn't separate them in different unit tests just to keep it DRY.

Also rebased with SeekHead and squashed to one commit. Of course, tests are passing.

lucab:
I didn't dig into it, but the test is failing for me locally:
```
--- FAIL: TestJournalCursorGetSeekAndTest (0.00s)
        journal_test.go:125: Error reading to journal
FAIL
```

glerchundi:
Umm, weird because I'm creating a new a `Journal` exactly in the same way as de `JournalReader` does but without matches and no kind of seeking (no since, no tail, no cursors).

The idea was to create the reader (which seeks internally near to the end of the journal), then write something into it and finally wait + read an entry (which could or couldn't be the one I wrote) in order to get the cursor.

I don't know why you're getting that error. I would expect an `Error waiting to journal` because it timed out  or something like that but not that one...

glerchundi:
@lucab I would like to reproduce your case, which is your environment? Directly CoreOS? I'm running the tests with go 1.5.1 linux/amd64 in a virtual machine with ubuntu 15.10 (which includes systemd as a default init system).

I re-run the test in a benchmark in case it was a unidentified race condition but no way...

lucab:
I saw this failure on a Debian testing amd64, with go 1.6.1 and systemd 229, building+testing from an unprivileged user session.

glerchundi:
@lucab i'm creating a new vm mimicking your environment but in the meantime can you re-run the test? I slightly modified to be a little bit more verbose. If we're lucky we'll get what's going on.

glerchundi:
@lucab reproduced and I firmly believe this is a permission issue:

Running as unprivileged user:
```
glerchundi@debian:~/.go/src/github.com/coreos/go-systemd$ ./test 
Running tests...
ok  	github.com/coreos/go-systemd/activation	5.456s	coverage: 0.0% of statements
?   	github.com/coreos/go-systemd/journal	[no test files]
ok  	github.com/coreos/go-systemd/login1	0.008s	coverage: 41.9% of statements
ok  	github.com/coreos/go-systemd/machine1	0.006s	coverage: 65.0% of statements
ok  	github.com/coreos/go-systemd/unit	0.004s	coverage: 93.7% of statements
--- FAIL: TestJournalCursorGetSeekAndTest (0.00s)
	journal_test.go:131: Error reading to journal: EOF
FAIL
coverage: 32.8% of statements
FAIL	github.com/coreos/go-systemd/sdjournal	5.870s
```

Running as root:
```
root@debian:/home/glerchundi/.go/src/github.com/coreos/go-systemd# ./test 
Running tests...
ok  	github.com/coreos/go-systemd/activation	6.073s	coverage: 0.0% of statements
?   	github.com/coreos/go-systemd/journal	[no test files]
ok  	github.com/coreos/go-systemd/login1	0.040s	coverage: 41.9% of statements
ok  	github.com/coreos/go-systemd/machine1	0.005s	coverage: 65.0% of statements
ok  	github.com/coreos/go-systemd/unit	0.003s	coverage: 93.7% of statements
ok  	github.com/coreos/go-systemd/sdjournal	7.480s	coverage: 41.3% of statements
ok  	github.com/coreos/go-systemd/dbus	2.188s	coverage: 62.1% of statements
Checking gofmt...
Checking govet...
Success
```

Digging a little deeper I checked if I've reading perms and as expected I haven't:
```
glerchundi@debian:~/.go/src/github.com/coreos/go-systemd$ journalctl -f
No journal files were found.
```

So in order to get the tests passing in your env add your unprivileged user to `systemd-journal` group or run it as `root`.

lucab:
You are right, this was a permission issue. Next() returned EOF in that case.

lucab:
LGTM then. I'll merge it for the moment, we can explore later if we relax test requirements later. @iaguis suggested OOB that it may be doable with a temporary journal and `NewJournalFromDir()`.

glerchundi:
Cool thanks @lucab!

> @iaguis suggested OOB that it may be doable with a temporary journal and NewJournalFromDir().

Seems reasonable, if I found a free slot I will give it a try.

sdjournal/read.go
```diff
@@ -33,6 +33,7 @@ type JournalReaderConfig struct {
 	// where the reading begins within the journal.
```
jonboulle: This comment should be updated to reflect that Cursor is also mutually exclusive (would also be good to list that they are tried in order)

glerchundi: @jonboulle yep, I missed it: /ref #163

PR #144, opened 2016-03-04
sdjournal: allow faster sd-journal api access with a no-op mutex.

glerchundi:


jonboulle:
before looking at the implementation in detail: can you provide more context? Why would someone want this? Any benchmarks for example?

glerchundi:
The idea is to use this lib for a journald log forwarder i'm working on. It should make the difference under high volume of journaling.

Here the benchmarker
```
$> cat mutex_test.go
package main

import (
  "sync"

  "testing"
)

func BenchmarkMutex(b *testing.B) {
    mu := &sync.Mutex{}
    for i := 0; i < b.N; i++ {
        mu.Lock()
        mu.Unlock()
    }
}

type noOpLocker struct {}

func (*noOpLocker) Lock() {}
func (*noOpLocker) Unlock() {}

func BenchmarkNoOpLocker(b *testing.B) {
    l := &noOpLocker{}
    for i := 0; i < b.N; i++ {
        l.Lock()
        l.Unlock()
    }
}
```

And the results:
```
$> go test -bench=. -benchtime=20s
testing: warning: no tests to run
PASS
BenchmarkMutex-8     	2000000000	        20.2 ns/op
BenchmarkNoOpLocker-8	2000000000	         0.30 ns/op
```

WDYT?

crawford:
A 20ns savings for an operation that is going to take many orders of magnitude longer doesn't seem worth the added complexity.

glerchundi:
Ok, closing this then. I though it was much more...

PR #147, opened 2016-04-08
unit: try to avoid unit file modification

kayrus:
@jonboulle it still skips commented strings, not sure how can I implement that...

main goal is to keep unit files unmodified when it is possible.

resolves https://github.com/coreos/fleet/issues/1539


jonboulle:
@kayrus I definitely appreciate the sentiment, but I'm a little ambivalent about this just being a partial promise, since as you note we still ignore comments. (personally, I regret the unit code ever went down the lexing path, since it's caused us so many problems :/).

kolyshkin:
This PR is very old, I think we should close it due to age. @kayrus if you still want to fix this, please open a new one. Sorry that it did not get much attention earlier.

PR #149, opened 2016-04-14
sdjournal: adds an async follower method

faiq:
this allows consumers to asynchronously receive values from journald. It seems like we don't need the polling logic that the regular follow method had. Would love to hear thoughts and suggestions!

### DIFF
--- a/sdjournal/read.go
+++ b/sdjournal/read.go
@@ -16,10 +16,12 @@
 package sdjournal
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"log"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


faiq:
This is good if you want to stream values from journald to a logging solution such as loggly.

sdjournal/read.go
```diff
@@ -188,6 +188,29 @@ process:
 	return
 }
 
+func (r *JournalReader) FollowAsync(until <-chan time.Time, recieve chan<- []byte) <-chan error {
```
jonboulle: few things:
- please add a docstring
- `s/recieve/receive`
- probably need to add a note that the `until` channel MUST be closed or this will leak a goroutine

sdjournal/read.go
```diff
@@ -188,6 +188,29 @@ process:
...
+	errCh := make(chan error, 1)
+	go func() {
+		var msg = make([]byte, 64*1<<(10))
+		for {
+			c, err := r.Read(msg)
+			if err != nil && err != io.EOF {
```
jonboulle: Hmm, you're still going to need to implement the logic @ https://github.com/coreos/go-systemd/blob/7b2428fec40033549c68f54e26e89e7ca9a9ce31/sdjournal/read.go#L156 , otherwise it's going to spin wildly at EOF again.

jonboulle: On thinking about this a bit more, we could probably just generalise the two cases and `Follow` could simply wrap around the async version

faiq:
@jonboulle thanks for getting back to me so fast! ended up implementing all of your suggestions and it looking pretty good to me. Let me know if there are other things you'd like me to change as well!

PR #150, opened 2016-04-21
dbus: add support for new ListUnits* methods

kayrus:
…d support)

is a part of this change: https://github.com/endocode/systemd/commit/5b4452a643d924015ed25972d071ad8802aad29c
and this pull request: https://github.com/coreos/fleet/pull/1564

### DIFF
--- a/dbus/methods.go
+++ b/dbus/methods.go
@@ -239,12 +239,11 @@ type UnitStatus struct {
 	JobPath     dbus.ObjectPath // The job object path
 }
 
-// ListUnits returns an array with all currently loaded units. Note that
-// units may be known by multiple names at the same time, and hence there might
-// be more unit names loaded than actual units behind them.
-func (c *Conn) ListUnits() ([]UnitStatus, error) {
+type storeFunc func(retvalues ...interface{}) error
+
+func (c *Conn) listUnitsInternal(f storeFunc) ([]UnitStatus, error) {
 	result := make([][]interface{}, 0)
-	err := c.sysobj.Call("org.freedesktop.systemd1.Manager.ListUnits", 0).Store(&result)
+	err := f(&result)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


dbus/methods.go
```diff
@@ -221,6 +221,36 @@ func (c *Conn) GetUnitTypeProperty(unit string, unitType string, propertyName st
 	return c.getProperty(unit, "org.freedesktop.systemd1."+unitType, propertyName)
 }
 
+// ListUnitsFiltered returns an array with units.
+// It takes a list of units' statuses and names to filter.
+// Note that units may be known by multiple names at the same time,
+// and hence there might be more unit names loaded than actual units behind them.
+func (c *Conn) ListUnitsFiltered(states []string, patterns []string) ([]UnitStatus, error) {
+	result := make([][]interface{}, 0)
+	err := c.sysobj.Call("org.freedesktop.systemd1.Manager.ListUnitsFiltered", 0, states, patterns).Store(&result)
```
jonboulle: is this documented somewhere? seems they've stopped updating the API page :-( https://www.freedesktop.org/wiki/Software/systemd/dbus/

kayrus: No, it is not, see https://github.com/systemd/systemd/issues/1042
Also this `ListUnitsFiltered` implementation is customized. I'm working on systemd patch as I mentioned in PR description.

kayrus:
@jonboulle updated this PR according to https://github.com/systemd/systemd/pull/3142

kayrus:
@jonboulle Updated this PR according to https://github.com/systemd/systemd/pull/3182

kayrus:
@jonboulle both new methods were merged in systemd.

dbus/methods.go
```diff
@@ -221,6 +221,66 @@ func (c *Conn) GetUnitTypeProperty(unit string, unitType string, propertyName st
...
+	return status, nil
+}
+
+// ListUnitsByNames returns an array with units. It takes a list of units'
+// names and returns an UnitStatus array. Comparing to ListUnitsByPatterns
+// method, this method returns statuses even for inactive or non-existing
+// units. Input array should contain exact unit names, but not patterns.
+func (c *Conn) ListUnitsByNames(units []string) ([]UnitStatus, error) {
+	result := make([][]interface{}, 0)
+	err := c.sysobj.Call("org.freedesktop.systemd1.Manager.ListUnitsByNames", 0, units).Store(&result)
```
jonboulle: This function is 99% the same as ListUnits, so can you split out the common code, something like

`type storeFunc func(retvalues ...interface{}) error`

```
func (c *Conn) listUnitsInternal(f storeFunc) ([]UnitStatus, error) {
	result := make([][]interface{}, 0)
	err := f(&result)
	if err != nil {
...  // rest of current func here
	return status, nil
}
```

```
func (c *Conn) ListUnitsByNames(units []string) ([]UnitStatus, error) {
  return c.listUnitsInternal(
     c.sysobj.Call("org.freedesktop.systemd1.Manager.ListUnitsByNames", 0, units).Store
  )
}
```

jonboulle:
LGTM in principle, but:
- [ ] style nit (for common code)
- [ ] any chance of some _some_ kind of testing? I realise it'll be difficult because this hasn't made it into a systemd release yet I guess..?

kayrus:
@jonboulle fixed PR following your suggestions. I'll provide benchmark results inside fleet PR: https://github.com/coreos/fleet/pull/1564

lucab:
Regarding previous review, code has been refactored but it is still not covered by any tests (I guess they can be added later, if fleet is blocked on this). Nothing else to add from my side.

kayrus:
@lucab does switching from `ListUnits` to `ListUnitsByNames` in `methods_test.go` sufficient?

lucab:
@kayrus I'm not sure if it is doable, but I would try to pick some well-known unit name/pattern (ie. existing on most systems) and try to test if the filtered listing succeeds and the result is a subset of the full set you get via `ListUnits`.

kayrus:
@lucab done

dbus/methods_test.go
```diff
@@ -130,6 +133,238 @@ func TestStartStopUnit(t *testing.T) {
...
+
+	unit = nil
+	for _, u := range units {
+		if u.Name == target2 {
+			unit = &u
+			break
+		}
+	}
+
+	if unit == nil {
```
jonboulle: Why should it be found in the list? this is counterintuitive the way it's written

kayrus: @jonboulle take a look on this comment https://github.com/coreos/go-systemd/pull/150/files#diff-b5b8384be79564335a3b50bcfed63e91R293
systemd should return unit states for non-existing units.

jonboulle: Oh... confusing

kayrus:
@lucab ping

dbus/methods_test.go
tixxdz: @jonboulle yep I confirm it's a bit confusing, but that's how systemd works actually please check this comment https://github.com/systemd/systemd/pull/3182#issuecomment-216828850 so basically the implementation doesn't bother it's just gives you back what you have requested setting up the right state of course.

Now this test can be updated to be more readable:
just create a target1 and target2 object which has {unit_name, activestate, a bolean found or not} , do the whole check inside the same loop, and check after the loop if the units have been found or not.  The doc of ListUnitsByNames() clearly states that it may return results for even non-existent units which is good.

dbus/methods_test.go
```diff
@@ -130,6 +133,238 @@ func TestStartStopUnit(t *testing.T) {
...
+	}
+
+	unit = nil
+	for _, u := range units {
+		if u.Name == target2 {
+			unit = &u
+			break
+		}
+	}
+
```
tixxdz: same thing with previous test ?

tixxdz:
@jonboulle 
> any chance of some some kind of testing? I realise it'll be difficult because this hasn't made it into a systemd release yet I guess..?

Not sure if it covers what you have in mind, but that change at systemd level was covered inside systemctl by a fallback to the old method if the new provided one method is not available. I think @kayrus did the right thing the check also fallbacks in fleet here:
https://github.com/coreos/fleet/pull/1564/files#diff-561bbb797eb2f941c1384271efe28c0bR208 

However to not have test failures in go-systemd, so inside newly added tests of go-systemd when you call the new methods, check the error code and if the method is not available do not fatal ? warning or something ? not sure if this make sense, it depends on how often go-systemd is updated, in all case systemd was released with these new changes.

Thanks!

tixxdz:
lgtm overall, just small code cleanups, thanks!

lucab:
@kayrus sorry for the delay. Except for @tixxdz comments above, I think it should be ready to go.

kayrus:
@lucab @jonboulle done

lucab:
LGTM

kayrus:
@lucab will you make some kind of tag?

lucab:
Yes, v9 is pending due to a regression in sdjournal.

kayrus:
@lucab please ping me when you'l fix that.

lucab:
Tagged v9 - https://github.com/coreos/go-systemd/releases/tag/v9

PR #151, opened 2016-04-25
sdjournal: adds a follow with fields

faiq:
this allows users to pull data from journal entries that aren't "MESSAGE"

would love to hear comments on this

### DIFF
--- a/sdjournal/read.go
+++ b/sdjournal/read.go
@@ -205,7 +205,7 @@ func (r *JournalReader) Follow(ctx context.Context, msgs chan<- map[string]inter
 					case <-pollDone:
 						return
 					default:
-						events <- r.journal.Wait(time.Duration(1) * time.Second)
+						events <- r.journal.Wait(time.Duration(300) * time.Second)
 					}
 				}
 			}()
--- a/sdjournal/read.go
+++ b/sdjournal/read.go
@@ -190,7 +190,10 @@ process:
 	return
 }
 
-func (r *JournalReader) FollowWithFields(until <-chan time.Time, fields map[string]struct{}, writer io.Writer) (err error) {
+// FollowWithFields synchronously follows the Journal, writing each full Journal Entry with all its given fields
+// as a map to the writer. Similar to Follow(), it will continue until a single time.Time is
+// received on the until channel.
+func (r *JournalReader) FollowWithFields(until <-chan time.Time, writer io.Writer) (err error) {
 	enc := json.NewEncoder(writer)
 journal:
 	for {
@@ -222,10 +225,7 @@ journal:
 					}
 					s = s[:len(s)]
 					arr := strings.SplitN(s, "=", 2)
-					if _, ok := fields[arr[0]]; ok {
-						kvMap[arr[0]] = arr[1]
-					}
-
+					kvMap[arr[0]] = arr[1]
 				}
 				if err := enc.Encode(kvMap); err != nil {
 					return err
--- a/sdjournal/read.go
+++ b/sdjournal/read.go
@@ -16,13 +16,14 @@
 package sdjournal
 
 import (
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"log"
 	"strings"
 	"time"
+
+	"golang.org/x/net/context"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


faiq:
again- could probably wrap `Follow` with `FollowWithFields`

sdjournal/read.go
```diff
@@ -188,6 +190,83 @@ process:
...
+			// get the data for what we care about and return
+			r.journal.RestartData()
+			if c > 0 {
+			fields:
+				for {
+					s, err := r.journal.EnumerateData()
+					if err != nil || len(s) == 0 {
+						break fields
+					}
+					s = s[:len(s)]
```
yifan-gu: Why we need this?

sdjournal/read.go
```diff
@@ -188,6 +190,83 @@ process:
...
+			// An unexpected error
+			if err != nil {
+				return err
+			}
+
+			// We have a new journal entry go over the fields
+			// get the data for what we care about and return
+			r.journal.RestartData()
+			if c > 0 {
+			fields:
```
yifan-gu: nit: maybe change to another name, or even better, put into a small function? Because `fields` is already taken an variable here, so it feels a little bit confusing.

sdjournal/read.go
faiq: Thanks for reviewing.

The `EnumerateData` call allows us to walk over all the different fields in the journal file. Now that I think about it, this can likely be done with multiple calls to `GetData` as well. Which would likely make it more efficient

sdjournal/read.go
faiq: Good idea.

sdjournal/read.go
```diff
@@ -188,6 +190,83 @@ process:
 	return
 }
 
+func (r *JournalReader) FollowWithFields(until <-chan time.Time, fields map[string]struct{}, writer io.Writer) (err error) {
```
yifan-gu: This is kinda copied from the Follow() function. But I feel it's better to just pass a time, then create a channel inside the function. (e.g. `untilCh := time.After(until)`)
cc @jonboulle ?

No need for a change in this PR though.

sdjournal/read.go
yifan-gu: @faiq Good idea.

sdjournal/read.go
```diff
@@ -188,6 +190,83 @@ process:
...
+		// reason to unblock the journal watch routine more quickly
+		events := make(chan int, 1)
+		pollDone := make(chan bool, 1)
+		go func() {
+			for {
+				select {
+				case <-pollDone:
+					return
+				default:
+					events <- r.journal.Wait(time.Duration(1) * time.Second)
```
yifan-gu: We need to refactor this waiting-polling logic in the future.
e.g. We should be able to pass a negative duration or some defined constant to journal.Wait() so it can wait infinitely.

From https://www.freedesktop.org/software/systemd/man/sd_journal_get_fd.html

> The maximum time this call sleeps may be controlled with the timeout_usec parameter. Pass (uint64_t) -1 to wait indefinitely.

sdjournal/read.go
```diff
@@ -188,6 +190,83 @@ process:
...
+				}
+			}
+		}()
+
+		select {
+		case <-until:
+			pollDone <- true
+			return ErrExpired
+		case e := <-events:
+			pollDone <- true
```
yifan-gu: @faiq I know this is copied from Follow(), but it doesn't feel right to me.
e.g. if there is no journal events in 1 second, then events r.journal.Wait() returns, and feed to the channel `events` in the above goroutine.
Then here, before we test the events, we notify the `pollDone` channel, causes the goroutine to return.

sdjournal/read.go
faiq: Actually in retrospect, i think this method should be called `FollowWithAllFields` because many people end up using custom fields in their journal entries. This allows the user to have the option to filter or not.

I think the best way of doing so would be taking out the `fields map[string]struct{}` parameter to the function

sdjournal/read.go
faiq: I like the idea of passing in the channel, because it allows me more flexibility when I write my programs with this library. For example, I am working on a daemon that forwards journald logs to an external logging service over http. The only reason I would ever want it killed is if a signal interrupt were given, so to do this I send a `time.Now()` from my signal handler into this channel to stop the program gracefully.

yifan-gu: @faiq Ok, sounds fair.

faiq:
closing in favor of #152

PR #152, opened 2016-04-27
Follow with filter

faiq:
Okay so this is a combination of my last 2 pull requests: #149 and #151 and I think this provides really what an end user would want. 

So let me describe the changes here: 

1) take in a ctx.Context rather than a time.Time

This allows more flexibility in use cases as described here: 

> For example, I am working on a daemon that forwards journald logs to an external logging service over http. The only reason I would ever want it killed is if a signal interrupt were given, so to do this I send a time.Now() from my signal handler into this channel to stop the program gracefully.

If I use a context, its more clear that I want to cancel processing of the logs rather than sending a time. Alternatively if someone wants to finish processing at a certain time, they can still do so. All they would have to do is send a ctx.Done() whenever the time hts

2) Use a filter function to figure out what part of the journald log you want

It's very reasonable for someone to want the entire journal entry, rather than just the message field. In fact, we use the Send() method from the journal library https://github.com/coreos/go-systemd/blob/master/journal/journal.go#L75 to write custom log fields for our programs into journald. It's only suitable that we should be able to pull them out for later processing

3) Send the journal entries with the filtered fields to `msgs chan<- map[string]interface{}`

This allows us to read all fields from the jorunal without worrying about the type and send them asynchronously without having to marshal and unmarshal the entries

Willing to close #151 in favor of this.

Thanks for all the help

### DIFF
--- a/sdjournal/read.go
+++ b/sdjournal/read.go
@@ -190,7 +190,10 @@ process:
 	return
 }
 
-func (r *JournalReader) FollowWithFields(until <-chan time.Time, fields map[string]struct{}, writer io.Writer) (err error) {
+// FollowWithFields synchronously follows the Journal, writing each full Journal Entry with all its given fields
+// as a map to the writer. Similar to Follow(), it will continue until a single time.Time is
+// received on the until channel.
+func (r *JournalReader) FollowWithFields(until <-chan time.Time, writer io.Writer) (err error) {
 	enc := json.NewEncoder(writer)
 journal:
 	for {
@@ -222,10 +225,7 @@ journal:
 					}
 					s = s[:len(s)]
 					arr := strings.SplitN(s, "=", 2)
-					if _, ok := fields[arr[0]]; ok {
-						kvMap[arr[0]] = arr[1]
-					}
-
+					kvMap[arr[0]] = arr[1]
 				}
 				if err := enc.Encode(kvMap); err != nil {
 					return err
--- a/sdjournal/read.go
+++ b/sdjournal/read.go
@@ -16,13 +16,14 @@
 package sdjournal
 
 import (
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"log"
 	"strings"
 	"time"
+
+	"golang.org/x/net/context"
 )
 
 var (
@@ -49,6 +50,22 @@ type JournalReaderConfig struct {
 // systemd journal.
 type JournalReader struct {
 	journal *Journal
+}
+
+// FollowFilter is a function which you pass into Follow to determine if you to
+// filter key value pair from a journal entry.
+type FollowFilter func(key, value string) bool
+
+// AllKeys is a FollowFilter that allows all keys of a given journal entry through the filter
+func AllKeys(key, value string) bool {
+	return true
+}
+
+func OnlyMessages(key, value string) bool {
+	if key == "MESSAGE" {
+		return true
+	}
+	return false
 }
 
 // NewJournalReader creates a new JournalReader with configuration options that are similar to the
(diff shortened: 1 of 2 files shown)


yifan-gu:
Sorry for not being responsive, (was not cc'd, so I lost this PR in notification).
Will take a review.

sdjournal/read.go
```diff
@@ -129,63 +148,85 @@ func (r *JournalReader) Close() error {
...
-			case SD_JOURNAL_NOP, SD_JOURNAL_APPEND, SD_JOURNAL_INVALIDATE:
-				// TODO: need to account for any of these?
-			default:
-				log.Printf("Received unknown event: %d\n", e)
+
+			// we're at the tail, so wait for new events or time out.
+			// holds journal events to process. tightly bounded for now unless there's a
+			// reason to unblock the journal watch routine more quickly
+			events := make(chan int, 1)
+			pollDone := make(chan bool, 1)
```
yifan-gu: Maybe change the size to 0, and close the channel instead of sending a bool?

sdjournal/read.go
```diff
@@ -129,63 +148,85 @@ func (r *JournalReader) Close() error {
...
+				if err != nil {
+					errChan <- err
 					return
-				default:
-					events <- r.journal.Wait(time.Duration(1) * time.Second)
+				}
+
+				// We have a new journal entry go over the fields
+				// get the data for what we care about and return
+				r.journal.RestartData()
```
yifan-gu: How about moving this below `if c > 0` ?

sdjournal/read.go
```diff
@@ -129,63 +148,85 @@ func (r *JournalReader) Close() error {
...
+				if c > 0 {
+				fields:
```
yifan-gu: Do we really need this tag?

sdjournal/read.go
```diff
@@ -129,63 +148,85 @@ func (r *JournalReader) Close() error {
...
-			return ErrExpired
-		case e := <-events:
-			pollDone <- true
-			switch e {
-			case SD_JOURNAL_NOP, SD_JOURNAL_APPEND, SD_JOURNAL_INVALIDATE:
-				// TODO: need to account for any of these?
-			default:
-				log.Printf("Received unknown event: %d\n", e)
+
+			// we're at the tail, so wait for new events or time out.
```
yifan-gu: I dont think this is true, we need to add a `continue` in the `if c > 0 { }`.
e.g.
```
if c > 0 {
    r.journal.RestartData()
    ...
    continue
}
```

A better way would be put all these into a function to make it cleaner
```
if c > 0 {
    readData()
    continue;
```

sdjournal/read.go
yifan-gu: Or better, can we reuse these channels ?

yifan-gu:
Ref https://github.com/coreos/rkt/pull/2041 as this is an API break change.

lucab:
@faiq ping :)

This stale PR has been pending for some time, do you still want to proceed on this? If so, can you please address @yifan-gu comments?

faiq:
yikes sorry about being so non-responsive and thanks for the ping. it turns out we did not end up needing this functionality. I am more than happy to finish work on this or close it, depending on if you think this patch would be helpful or not.

sdjournal/read.go
```diff
@@ -129,63 +148,85 @@ func (r *JournalReader) Close() error {
...
-		if err != nil && err != io.EOF {
-			break process
-		}
+// Follow asynchronously follows the JournalReader it takes in a context to stop the following the Journal, it returns a
+// buffered channel of errors and will stop following the journal on the first given error
+func (r *JournalReader) Follow(ctx context.Context, msgs chan<- map[string]interface{}, filter FollowFilter) <-chan error {
+	errChan := make(chan error, 1)
+	go func() {
+		for {
+			kvMap := make(map[string]interface{})
```
lucab: Why value type is `interface{}`? Aren't string enough?

lucab:
@faiq I think it would be nice to have some portions of this.

Point (1) looks good and point (3) makes the `Follow()` API more useful for the general case (even though I'm not sure why values are interfaces). We now have a `GetEntry()` which should help reducing the size of this PR.

Regarding point (2) I think the library shouldn't be concerned about this. And once the other two points are in place, it should be easy for consumers to just provide their own filtering wrapper.

faiq:
Sorry I'm so slow at responding to this 😞 lots of work makes it hard to contribute to open source.

regarding why we're sending `interface{}` through the channel, this is because the values on the fields can be json by something like https://github.com/wercker/journalhook and logrus. 

i can probably take out the enumerating logic for a call to `GetEntry` instead, and filter in go-land.

how does this all sound?

lucab:
> regarding why we're sending interface{} through the channel, this is because the values on the fields can be json by something like https://github.com/wercker/journalhook and logrus.

I'm not completely sure I'm understanding your proposal. Current API will return you either a `string` or a `[]byte`, so even if the inner content is a JSON the receiver still need to unmarshal it. E.g. in the case of `GetEntry` below, this is already a `map[string]string`.

> i can probably take out the enumerating logic for a call to GetEntry instead, and filter in go-land.

Just to be clear, who will be performing the filtering? Do you need go-systemd to filter fields, or will the receiver take care of that? I was suggesting the latter, but I may understand in some performance-related scenarios the former may be interesting too...

kolyshkin:
In general, I like the direction of these changes, but the PR itself is way too old to revive, and there are some API incompatibilities (e.g. Follow should be kept as it, and maybe deprecated in favor of a new `FollowContext`).

@faiq I apologize that this wasn't finished in time, but if you still want to work on it, please feel free to open a new PR.

PR #153, opened 2016-04-28
Expose dbus.NewConnection

nathan7:
I'm trying to use the go-systemd/dbus package with DBus sockets established over SSH.
This is mostly a straw PR, since we could probably find a better way of exposing this.

### DIFF
--- a/dbus/dbus.go
+++ b/dbus/dbus.go
@@ -86,7 +86,7 @@ type Conn struct {
 // New establishes a connection to the system bus and authenticates.
 // Callers should call Close() when done with the connection.
 func New() (*Conn, error) {
-	return newConnection(func() (*dbus.Conn, error) {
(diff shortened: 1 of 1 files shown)


jonboulle:
Could you provide a little more detail on how you set this up? What's your code look like at the moment? I imagine we can make this easier by exposing a transport or something

nathan7:
Right now, my transport implements `interface{ DBus() (*dbus.Conn, error) }`, and passing that to `NewConnection` does the trick. (https://godoc.org/go.nathan7.eu/nyb#Client)
Under the hood, that either connects to the local system bus, or a remote one using SSH domain socket forwarding.

jonboulle:
This change looks reasonable, but could you add a docstring describing the func and how people might use it?

nathan7:
@jonboulle How's this?

PR #157, opened 2016-05-11
Add // +build ignore to examples

mwhudson:
This means that go get github.com/coreos/go-systemd/... can succeed.


natefinch:
Nice.  In case it's not obvious, the build tags get ignored if you're using go run foo.go, which presumably is how these examples are intended to be run.

yifan-gu:
LGTM

mwhudson:
Belated thanks, not sure how I failed to notice this being merged!

PR #159, opened 2016-05-19
Add test for dbus/methods.go:GetServiceProperty function

brnv:


lucab:
Great, :+1: 
Thanks a lot for the follow-up!

PR #161, opened 2016-05-26
sdjournal: export SYSLOG_IDENTIFIER

iaguis:
rkt needs this because there's a race where short-lived non-root
processes make the journal unable to get all their information (i.e. the
SYSTEMD_UNIT). We work around it by using SYSLOG_IDENTIFIER.


jonboulle:
lgtm

PR #164, opened 2016-05-26
sdjournal: add GetEntry method to retrieve all fields

### LINKED ISSUE #142: sdjournal: Enumerate the fields
Right now, `sdjournal` only lets you access fields you already know to expect. The C library has `sd_journal_enumerate_data` for accessing all fields of a single message, and `sd_journal_enumerate_fields` for listing all fields that all messages in the journal contain.

https://www.freedesktop.org/software/systemd/man/sd_journal_get_data.html
https://www.freedesktop.org/software/systemd/man/sd_journal_enumerate_fields.html

glerchundi:
Implements the same semantics as the JOURNAL_FOREACH_DATA_RETVAL macro in journal-internal.h and produces a *JournalEntry with every field with its corresponding data in a map[string]string. Also it includes address fields like realtime (__REALTIME_TIMESTAMP), monotonic (__MONOTONIC_TIMESTAMP and cursor (__CURSOR).

It mimics more or less `journalctl -o json`.

Fixes #142
Supersedes #136


wrouesnel:
Looks good to me : should it support JsonEncode to produce a matching output?

glerchundi:
@wrouesnel sounds reasonable but I would vote not to do so. I think json encoding is out of this package scope (although in my use-case would be a really helpful shortcut...)

WDYT on this @lucab @jonboulle?

lucab:
Personally, I also think json encoding is out of scope here and easily doable in the consuming application. (I just had a quick glance at code, haven't reviewed yet)

glerchundi:
> (I just had a quick glance at code, haven't reviewed yet)

Ok, thanks!

sdjournal/journal_test.go
```diff
@@ -150,3 +151,47 @@ func TestJournalCursorGetSeekAndTest(t *testing.T) {
...
+	entry, err := j.GetEntry()
+	if err != nil {
+		t.Fatalf("Error getting the entry to journal: %s", err)
+	}
+
+	got := entry.Fields["MESSAGE"]
+	if got != want {
+		t.Fatalf("Bad result for entry.Fields[\"MESSAGE\"]: got %s, want %s", got, want)
+	}
+}
```
lucab: This test is fragile, it will easily pick up messages from other tests and get confused. I think you should instead filter on something a bit more specific (for example introducing a custom field key with a random entry, and matching on that).

sdjournal/journal.go
```diff
@@ -245,15 +273,18 @@ var libsystemdFunctions = map[string]unsafe.Pointer{}
...
-	SD_JOURNAL_FIELD_SYSLOG_IDENTIFIER = "SYSLOG_IDENTIFIER"
-	SD_JOURNAL_FIELD_MESSAGE           = "MESSAGE"
-	SD_JOURNAL_FIELD_PID               = "_PID"
-	SD_JOURNAL_FIELD_UID               = "_UID"
-	SD_JOURNAL_FIELD_GID               = "_GID"
-	SD_JOURNAL_FIELD_HOSTNAME          = "_HOSTNAME"
-	SD_JOURNAL_FIELD_MACHINE_ID        = "_MACHINE_ID"
-	SD_JOURNAL_FIELD_TRANSPORT         = "_TRANSPORT"
+	SD_JOURNAL_FIELD_SYSTEMD_UNIT        = "_SYSTEMD_UNIT"
+	SD_JOURNAL_FIELD_SYSLOG_IDENTIFIER   = "SYSLOG_IDENTIFIER"
```
lucab: While at it, I think we may benefit for a couple more fields here like: `PRIORITY`, all `SYSLOG_*` ones, all `SYSTEMD_*` ones. And please split it in a separate commit.

sdjournal/journal.go
```diff
@@ -583,6 +622,93 @@ func (j *Journal) GetDataValue(field string) (string, error) {
...
+	var l C.size_t
+	C.my_sd_journal_restart_data(sd_journal_restart_data, j.cjournal)
+	for {
+		r = C.my_sd_journal_enumerate_data(sd_journal_enumerate_data, j.cjournal, &d, &l)
+		if r == 0 {
+			break
+		}
+
+		if r < 0 {
+			return nil, fmt.Errorf("failed to read message field: %d", r)
```
lucab: Here too, `r` is an errno-style code.

lucab:
Some remarks inline. I think the test definitely needs to be reworked, I already saw it failing several times here (it is a pity this can't be run on Travis).

sdjournal/journal.go
glerchundi: What do you mean? Ease access to other non address fields by resolving them from key values instead of a generic one? Something like:

```
// JournalEntry represents all fields of a journal entry plus address fields.
type JournalEntry struct {
	Fields             map[string]string
	Priority           string?
	Cursor             string
	RealtimeTimestamp  uint64
	MonotonicTimestamp uint64
}
```

WDYT if we leave this for another pull request? If we agree I'll work on that after this one.

sdjournal/journal.go
glerchundi: ups, i missed that, thanks.

sdjournal/journal.go
lucab: No, sorry if it was unclear, I simply meant exposing some more const identifiers here for fields that users would typically manipulate (not touching `JournalEntry`). It is not directly in-scope of your PR, so as you wish.

lucab:
@ wrouesnel @wolfeidau: as original supporters of this, do you have any other concerns with it? Does it cover your usecases (minus the json part)?

wrouesnel:
None from me!

sdjournal/journal_test.go
tv42: Maybe filter by `_PID`.

glerchundi:
@lucab PTAL, i fixed your comments and reworked the test using a custom field with a timestamp as a random string, WDYT?

lucab:
@glerchundi Thanks, I think it is much better now. I'll take another pass at this, just to double-check the C-casting part. In the meanwhile, would you mind word-wrapping the commit message as described in the [contributing page](https://github.com/coreos/go-systemd/blob/master/CONTRIBUTING.md#format-of-the-commit-message) and add a "fixes" line for #142?

glerchundi:
@lucab done, I referenced the issue in the commit, PTAL whenever you have time, thanks :)

glerchundi:
I reread every comments and splitted this PR in two commits, one for the implementation itself and the other for facilitating access to the fields themselves.

lucab:
I think you just hit an unrelated bug here, as most of the time the test fails for me with:
```
--- FAIL: TestJournalGetEntry (0.00s)
        journal_test.go:188: Error reading to journal: EOF
```

Hardcoding the match filter works, though, so I'm wondering if the bug lies somewhere else in the match subsystem. Is the test working fine on your system? Which systemd version are you running?

glerchundi:
@lucab wow, that's weird, I run the test against two environments, Debian and Ubuntu (both inside virtual machines):

```
glerchundi@debian:~/.go/src/github.com/coreos/go-systemd$ ./test 
Running tests...
ok  	github.com/coreos/go-systemd/activation	6.788s	coverage: 0.0% of statements
?   	github.com/coreos/go-systemd/journal	[no test files]
ok  	github.com/coreos/go-systemd/login1	0.005s	coverage: 41.9% of statements
ok  	github.com/coreos/go-systemd/machine1	0.006s	coverage: 65.0% of statements
ok  	github.com/coreos/go-systemd/unit	0.003s	coverage: 93.7% of statements
ok  	github.com/coreos/go-systemd/sdjournal	6.948s	coverage: 45.2% of statements
Checking gofmt...
Checking govet...
Success

glerchundi@debian:~/.go/src/github.com/coreos/go-systemd$ uname -a
Linux debian 3.16.0-4-amd64 #1 SMP Debian 3.16.7-ckt25-2 (2016-04-08) x86_64 GNU/Linux

glerchundi@debian:~/.go/src/github.com/coreos/go-systemd$ cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 8 (jessie)"
NAME="Debian GNU/Linux"
VERSION_ID="8"
VERSION="8 (jessie)"
ID=debian
HOME_URL="http://www.debian.org/"
SUPPORT_URL="http://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"

glerchundi@debian:~/.go/src/github.com/coreos/go-systemd$ systemctl --version
systemd 215
+PAM +AUDIT +SELINUX +IMA +SYSVINIT +LIBCRYPTSETUP +GCRYPT +ACL +XZ -SECCOMP -APPARMOR
```

```
glerchundi@ubuntu:~/.go/src/github.com/coreos/go-systemd$ ./test
Running tests...
ok  	github.com/coreos/go-systemd/activation	5.222s	coverage: 0.0% of statements
?   	github.com/coreos/go-systemd/journal	[no test files]
ok  	github.com/coreos/go-systemd/login1	0.005s	coverage: 41.9% of statements
ok  	github.com/coreos/go-systemd/machine1	0.009s	coverage: 65.0% of statements
ok  	github.com/coreos/go-systemd/unit	0.006s	coverage: 93.7% of statements
ok  	github.com/coreos/go-systemd/sdjournal	6.490s	coverage: 45.0% of statements
Checking gofmt...
Checking govet...
Success

glerchundi@ubuntu:~/.go/src/github.com/coreos/go-systemd$ uname -a
Linux ubuntu 4.2.0-30-generic #36-Ubuntu SMP Fri Feb 26 00:58:07 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux

glerchundi@ubuntu:~/.go/src/github.com/coreos/go-systemd$ cat /etc/os-release 
NAME="Ubuntu"
VERSION="15.10 (Wily Werewolf)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 15.10"
VERSION_ID="15.10"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"

glerchundi@ubuntu:~/.go/src/github.com/coreos/go-systemd$ systemctl --version
systemd 225
+PAM +AUDIT +SELINUX +IMA +APPARMOR +SMACK +SYSVINIT +UTMP +LIBCRYPTSETUP +GCRYPT +GNUTLS +ACL +XZ -LZ4 +SECCOMP +BLKID -ELFUTILS +KMOD -IDN
```

And it worked on both...., which systemd version are you using?

glerchundi:
@lucab can you try if doing some retries, in this case 5 (I made this in the same way your reading follower does), works on your machine? Maybe journald is not emitting the result that fast to the sd_* subsystem and we're facing a race condition here.

lucab:
@glerchundi yes I think I was hitting the R/W race you describe, just retrying one more time seems to reliably pass the test. I have no better proposals here, so I'll merge it as is.
LGTM

glerchundi:
@lucab ok, wait a minute I slightly changed just for your use case, i'll rever to the last time you reviewed.

lucab:
@glerchundi ah, no hurry then. I actually think you can leave the retry there, as is the test is inherently racy.

glerchundi:
Ah, I reverted already because I'm quite sure once the EOF (r==0) is emitted next wait will fail so if you don't care I think it's better to keep it as simple as possible and think in a better solution, WDYT?

glerchundi:
@lucab should be mergeable now, recleaned commits, squashed, rebased and that git hell.

lucab:
@glerchundi sorry for the confusion :smile:. Yes, it is ok to land this as is and then explore other issues/races separately. I just double-checked the commits in this branch and they seem ok now. May I proceed?

glerchundi:
Yes please!

glerchundi:
🎉 🎉

Thanks for your time @lucab

PR #166, opened 2016-06-03
sdjournal: fix error mgmt in NewJournalFromDir

alban:
Errors were previously hidden. With this patch, I can see the following
error:

> $ ./client_example
> rpc error: code = 2 desc = "error resolving symbol \"sd_journal_next\": .2: undefined symbol: sd_journal_next"

Reproducible steps on https://github.com/coreos/rkt/issues/2740. This is
not solving the symbol issue, but at least, the error is not hidden.

### DIFF
--- a/sdjournal/journal.go
+++ b/sdjournal/journal.go
@@ -323,12 +323,13 @@ func NewJournal() (j *Journal, err error) {
 		return nil, err
 	}
 	defer func() {
-		if err == nil {
-			return
-		}
 		err2 := h.Close()
 		if err2 != nil {
-			err = fmt.Errorf(`%q and "error closing handle: %v"`, err, err2)
+			if err != nil {
+				err = fmt.Errorf(`%q and "error closing handle: %v"`, err, err2)
+			} else {
+				err = fmt.Errorf(`error closing handle: %v`, err2)
+			}
 		}
 	}()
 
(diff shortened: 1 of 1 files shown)


sdjournal/journal.go
```diff
@@ -357,7 +357,7 @@ func NewJournalFromDir(path string) (j *Journal, err error) {
 		return nil, err
 	}
 	defer func() {
-		if err == nil {
+		if err != nil {
```
jonboulle: surely you still want to close the handle in this case?

lucab: I think this was meant to be a resource cleaner on error, and this is now subverting the logic. OTOH this snippet looks over-complex, and I think the root issue here is implicitly reusing `err` for multiple purposes.

alban: right. Oh, defer functions and implicit returns can be confusing. I'll fix this.

alban:
Patch updated.

sdjournal/journal.go
```diff
@@ -323,12 +323,13 @@ func NewJournal() (j *Journal, err error) {
...
-			return
-		}
 		err2 := h.Close()
 		if err2 != nil {
-			err = fmt.Errorf(`%q and "error closing handle: %v"`, err, err2)
+			if err != nil {
+				err = fmt.Errorf(`%q and "error closing handle: %v"`, err, err2)
+			} else {
+				err = fmt.Errorf(`error closing handle: %v`, err2)
+			}
```
lucab: I fear this is still not correct. If I'm not wrong, this now always closes the dlopen handle (even if there are no errors) which is later used by several Journal methods.

lucab:
@alban There is something I don't understand here. You say this unmasks some errors, but then the error metions `sd_journal_next` which is not even called here. What if, instead, your patch prematurely closes the dlopened handle when there are no errors, which then triggers an error on the next method call? As in, perhaps you are introducing a bug here instead of fixing it?

alban:
@lucab I think you're right. With https://github.com/coreos/rkt/pull/2744 and https://github.com/coreos/rkt/pull/2747, I cannot reproduce the crash.

Issue #165, opened 2016-05-31
test: unit tests fail on ubuntu yakkety


kayrus:
```sh
sudo ./test 
Running tests...
ok      github.com/coreos/go-systemd/activation 4.550s  coverage: 0.0% of statements
?       github.com/coreos/go-systemd/journal    [no test files]
ok      github.com/coreos/go-systemd/login1     0.003s  coverage: 41.9% of statements
ok      github.com/coreos/go-systemd/machine1   0.003s  coverage: 65.0% of statements
ok      github.com/coreos/go-systemd/unit       0.002s  coverage: 93.7% of statements
ok      github.com/coreos/go-systemd/sdjournal  5.279s  coverage: 36.7% of statements
--- FAIL: TestEnableDisableUnit (0.00s)
        methods_test.go:353: Invalid argument
--- FAIL: TestGetUnitProperties (0.00s)
        methods_test.go:382: / is unwanted
--- FAIL: TestSetUnitProperties (0.00s)
        methods_test.go:454: Unit tmp.mount not found.
FAIL
coverage: 63.1% of statements
FAIL    github.com/coreos/go-systemd/dbus       2.143s
```

lucab:
I also see two more failures on a Debian stretch (systemd v230):
```
--- FAIL: TestStartStopUnit (0.03s)
        methods_test.go:107: Test unit not active
--- FAIL: TestEnableDisableUnit (0.00s)
        methods_test.go:163: Invalid argument
--- FAIL: TestGetUnitProperties (0.00s)
        methods_test.go:192: / is unwanted
--- FAIL: TestSetUnitProperties (0.00s)
        methods_test.go:264: Unit tmp.mount not found.
--- FAIL: TestStartStopTransientUnit (0.01s)
        methods_test.go:313: Test unit not active
FAIL
coverage: 58.0% of statements
FAIL    github.com/coreos/go-systemd/dbus       2.091s
```

kayrus:
Latest coreos alpha:

```sh
--- FAIL: TestEnableDisableUnit (0.04s)
        methods_test.go:326: Invalid argument
--- FAIL: TestGetUnitProperties (0.00s)
        methods_test.go:355: / is unwanted
```

lucab:
[...]

PR #168, opened 2016-06-06
dbus: add PropPids to set initial pids of transient scope units

gebi:
Should there also be a usage example on how to create a transient scope unit instead of a transient service unit?

Feel free to use https://github.com/mgit-at/systemd-runas/blob/master/main.go as a example on how to use the call with transient scopes.

### DIFF
--- a/dbus/methods_test.go
+++ b/dbus/methods_test.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
+	"os/exec"
 	"path"
 	"path/filepath"
 	"reflect"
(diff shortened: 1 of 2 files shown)


lucab:
Thanks for this PR!

> Should there also be a usage example on how to create a transient scope unit instead of a transient service unit?

Examples are always welcome, even though in the specific case I'm not sure where it would fit, do you have any specific idea in mind? I was actually thinking if we could use it for augmenting this PR with a test (a basic "create transient scope with these pids and check the properties"), without getting too much into muddy waters.

gebi:
Hi @lucab, thx :)!

hm... as for travis-ci, the dbus section is not included in the tests because it requires root to run, i'm not sure how those tests are being run and where.

lucab:
You are right, there are some tests which are not run by CI (and that's why they are in a [sub-optimal state](https://github.com/coreos/go-systemd/issues/165)), but nonetheless we try to keep them updated and use them to sanity check releases.

lucab:
@gebi ping :)

I've just queued up #178 and #179, which should fix current dbus testsuite and let it run on travis. How do we want to proceed here with your PR? Is it ok and do you have time to add a testcase for it, as proposed above?

lucab:
Hi @gebi, testsuite rework landed and travis is now excercising also the dbus package. This PR has been pending for some weeks, do you think you have time to attach a test to it?

gebi:
Hi @lucab 

oh, that is nice, thx for travis integration and tests of this part, makes the whole thing a _lot_ easier!

With this i should get this going, i'll try to add a meaning full test as soon as i have a bit of time.

gebi:
@lucab can i rebase this PR to current master to get the benefit of your #178 and #179?

lucab:
@gebi yes please! Once you rebase and push, travis should pick it up.

gebi:
@lucab is this unit test good enough?

lucab:
I just have a bunch of doubts:
 * why is it spinning an additional sleeping process? Isn't `os.GetPid()` on the original test process enough?
 * I saw your final remark, maybe we can at least check if the property has been applied via `GetUnitProperties()`?

gebi:
I have choose the route with an additional sleep process because i didn't want to tamper with the main process regarding cgroups as this might influence other unit tests further down the line.

As for `GetUnitProperties()`, i've tried adding them to the unit test, but i get "Unknown property or interface" when asking for "PIDs".

lucab:
I see the test failure, I'll try to have a look at what's going on.

lucab:
@gebi I finally had a look at it, and this can't work as we are not supposed to get PIDs via properties.

Instead, systemd exposes a `GetProcesses()` [dbus method](https://github.com/systemd/systemd/blob/b27b4b51c6fa3963941da3b65312384be1aedd69/src/core/dbus-unit.c#L993) for units, which I think is the one that should be used here. However we are currently not wrapping it in go-systemd. For the moment for this PR I'm thus ok if you just leave a TODO note in the last part of the test mentioning it. We will close the loop once `GetProcesses()` is implemented.

gebi:
Sounds good, i've removed the last commit where the add additional checks where added.

lucab:
Thanks, LGTM!

PR #169, opened 2016-06-07
sdjournal: tie libsystemdFunctions to Journal's life

iaguis:
There was a map from function names to their addresses so we don't
need to call dlsym everytime we call a function from libsystemd if it
was already called before.

Every time we open a new Journal, we call dlopen() (and every time we
close it, we call dlclose()). Since the map was global and we weren't
clearing it up after closing Journal, we were reusing the addresses from
the previous invocation. This is wrong because libsystemd can end up in
a different address the second time we dlopen() it.

Store the map in the Journal struct so it's a per-Journal cache.

Should fix https://github.com/coreos/rkt/issues/2740

### DIFF
--- a/sdjournal/journal.go
+++ b/sdjournal/journal.go
@@ -240,8 +240,6 @@ import (
 	"github.com/coreos/pkg/dlopen"
 )
 
-var libsystemdFunctions = map[string]unsafe.Pointer{}
-
 // Journal entry field strings which correspond to:
 // http://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html
 const (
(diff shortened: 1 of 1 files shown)


krnowak:
LGTM.

alban:
LGTM if Semaphore is green.

As said OOB, it's probably better to keep the library open rather than doing dlopen/dlclose at each method call. But this PR is a small patch that we can easily get in for the next release.

iaguis:
> dlopen/dlclose at each method call

FWIW we only do it when the user creates a new Journal (and when she closes it).

lucab:
I'm not sure this is a great approach: memory footprint will now be linear in the number of journal instances, and every instance will go through several misses every time (to populate symbols cache). Booking overhead, in particular on short lived instance, seems considerable now. Can't we just dlopen it once and then let it go away on process exit, reusing the same symbols cache for all instances?

iaguis:
This PR implements the suggestion on https://github.com/coreos/go-systemd/pull/169#issuecomment-224296087: https://github.com/coreos/go-systemd/pull/170

lucab:
Superseded by #170.

Issue #172, opened 2016-06-11
JournalReader.Read() doesn't return the correct number of bytes that's been read


yifan-gu:
Here in [JournalReader.Read()](https://github.com/coreos/go-systemd/blob/master/sdjournal/read.go#L132), the returned length can be larger than what's actually been read when the buffer provided is not big enough.
Which will cause a panic in https://golang.org/src/bytes/buffer.go#L177. (go 1.6.2 linux/amd64)

I found this when doing `kubectl logs on some pods`.

Here is the panic trace from the rkt api service:
```
Jun 11 01:05:48 yifan-test-rkt-minion-group-fc7i rkt[1212]: panic: runtime error: slice bounds out of range [recovered]
Jun 11 01:05:48 yifan-test-rkt-minion-group-fc7i rkt[1212]:         panic: runtime error: slice bounds out of range
Jun 11 01:05:48 yifan-test-rkt-minion-group-fc7i rkt[1212]: goroutine 6580 [running]:
Jun 11 01:05:48 yifan-test-rkt-minion-group-fc7i rkt[1212]: panic(0xdbd0c0, 0xc82000c030)
Jun 11 01:05:48 yifan-test-rkt-minion-group-fc7i rkt[1212]:         /usr/lib/go-1.6/src/runtime/panic.go:481 +0x3e6
Jun 11 01:05:48 yifan-test-rkt-minion-group-fc7i rkt[1212]: io/ioutil.readAll.func1(0xc8201b9950)
Jun 11 01:05:48 yifan-test-rkt-minion-group-fc7i rkt[1212]:         /usr/lib/go-1.6/src/io/ioutil/ioutil.go:30 +0x11e
Jun 11 01:05:48 yifan-test-rkt-minion-group-fc7i rkt[1212]: panic(0xdbd0c0, 0xc82000c030)
Jun 11 01:05:48 yifan-test-rkt-minion-group-fc7i rkt[1212]:         /usr/lib/go-1.6/src/runtime/panic.go:443 +0x4e9
Jun 11 01:05:48 yifan-test-rkt-minion-group-fc7i rkt[1212]: bytes.(*Buffer).ReadFrom(0xc8201b98a8, 0x7f9d80c0b780, 0xc82012c428, 0x0, 0x0, 0x0)
Jun 11 01:05:48 yifan-test-rkt-minion-group-fc7i rkt[1212]:         /usr/lib/go-1.6/src/bytes/buffer.go:177 +0x322
[...]

PR #173, opened 2016-06-13
sdjournal: fix incomplete reads in JournalReader.Read()

alban:
This patch fixes two issues in JournalReader.Read():

1. The return value of copy() was ignored and JournalReader.Read() was
returning the wrong amount of bytes copied.

2. JournalReader implements the Reader interface so it must be able to read a
specific amount of bytes. Journald on the other hand only allows us to read
full entries of arbitrary size (without byte granularity). JournalReader
therefore needs to internally buffer entries that don't fit in the read buffer.
Callers should keep calling until 0 and/or an error is returned.

Fixes https://github.com/coreos/go-systemd/issues/172

-----

/cc @yifan-gu

### DIFF
--- a/sdjournal/read.go
+++ b/sdjournal/read.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"strings"
 	"time"
 )
 
@@ -46,9 +47,10 @@ type JournalReaderConfig struct {
 }
 
 // JournalReader is an io.ReadCloser which provides a simple interface for iterating through the
-// systemd journal.
+// systemd journal. A JournalReader is not safe for concurrent use by multiple goroutines.
 type JournalReader struct {
-	journal *Journal
+	journal   *Journal
+	msgReader *strings.Reader
 }
 
 // NewJournalReader creates a new JournalReader with configuration options that are similar to the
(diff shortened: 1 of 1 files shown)


alban:
I am testing with:
```
sudo rkt run docker://busybox --exec sh -- -c 'while sleep 1 ; do hexdump /bin/busybox ; done'
sudo rkt api-service
./client_example -follow
```

No crashes, but the client does not follow the logs correctly.

sdjournal/read.go
```diff
@@ -103,33 +106,42 @@ func NewJournalReader(config JournalReaderConfig) (*JournalReader, error) {
...
+		}
+		r.msgReader = strings.NewReader(msg)
 	}
 
 	// Copy and return the message
-	copy(b, []byte(msg))
+	var sz int
+	sz, err = r.msgReader.Read(b)
+	if sz == r.msgReader.Len() {
+		r.msgReader = nil
```
yifan-gu: @alban this seems to generate some gc pressure. Not sure if that's going to be a big deal.
Wonder if we can update the string instead of creating the new reader struct every time we read?

alban:
Patch updated, PTAL.

Sometimes, I have a 30 seconds delay before getting new entries from the journal, but I think that's a problem in journald, depending on how fast it syncs the data on disk.

sdjournal/read.go
alban: Did you notice the gc pressure? If so, how did you produce it?

yifan-gu:
LGTM

lucab:
I have a probably naïf question, so sorry for asking: if this is to fix a wrong size return, wouldn't using copy()'s return-value be enough? Why introducing an additional strings.Reader? Maybe it fixes something else which is non-obvious to me...

yifan-gu:
@lucab because the copy() will not copy all the bytes if the provided `b []byte` is smaller than the string we read from.
Than how do we keep the remaining string?

alban:
I should have updated the commit message but I didn't, sorry. `r.buildMessage()` gets one full entry from journald. The cursor in journald points to an entry but is not granular to the byte so we must keep some context for the Golang Reader.

jonboulle:
It scares me a bit for the `JournalReader` to be maintaining this state between calls (consider `Read(); Rewind(); Read()`). At the very least I think `Read()` needs a docstring explaining very clearly that it's doing this buffering and callers must keep calling until 0 and/or an error is returned

lucab:
Ah, I didn't spot the change of semantics at first, my bad. It looks like this is now conflating two cases into EOF-returning (Next's EOF and log entry's EOL) but I'm not sure what (if any) will be the effect on applications.

alban:
I missed the issue with `Rewind()`. I updated the patch to reset the buffering in `Rewind()` and added documentation in `Read()` and updated the commit message.

@lucab I think `JournalReader.Read()` must not return EOF for the current log entry, otherwise applications would think this is an EOF for the file. It should never happen because I reset the string reader when `r.msgReader.Len() == 0`, unless we have a journal entry of zero bytes in the first place. Would it be better if I panic or just ignore the error?
```
	sz, err = r.msgReader.Read(b)
	if err != nil {
		panic("journal entry of zero bytes. This should not happen")
	}
```
or
```
	sz, _ = r.msgReader.Read(b)
```

sdjournal/read.go
```diff
@@ -48,7 +49,8 @@ type JournalReaderConfig struct {
 // JournalReader is an io.ReadCloser which provides a simple interface for iterating through the
 // systemd journal.
 type JournalReader struct {
```
jonboulle: A JournalReader is not safe for concurrent use by multiple goroutines.

sdjournal/read.go
```diff
@@ -101,35 +103,53 @@ func NewJournalReader(config JournalReaderConfig) (*JournalReader, error) {
...
+		}
+		r.msgReader = strings.NewReader(msg)
 	}
 
 	// Copy and return the message
-	copy(b, []byte(msg))
+	var sz int
+	sz, err = r.msgReader.Read(b)
+	if err != nil {
+		return sz, err
```
lucab: This may return (sz > 0, EOF) once the buffer is completely consumed, at least in theory according to Reader interface.

lucab:
> It should never happen

I left a comment inline pinpointing what I was referring to. I'm not sure if it can really happen, but I got that impression while checking the interface doc.

> Would it be better if I panic or just ignore the error?

I generally prefer to avoid panics in libraries and have typed error wherever possible, but I know this is just my personal taste. In this case it should probably special case the EOF error condition, and clean the reader + bubble-up the error in other cases.

alban:
@jonboulle @lucab Thanks for the review. Patch updated.

euank:
This code looks dangerous and I'd like to see tests for it (though they can be in a follow up issue if needed)... but afaict this looks right and resolves the original issue, so LGTM.

lucab:
Also LGTM. For double-checking I added a testcase for this, up at #182; it [tested fine](https://travis-ci.org/lucab/go-systemd/builds/138901851) on travis.

PR #174, opened 2016-06-13
sdjournal: Break the loop if error happens during Write().

yifan-gu:
This fixes the endless loop issue reported by @alban .

### DIFF
--- a/sdjournal/read.go
+++ b/sdjournal/read.go
@@ -161,7 +161,9 @@ process:
 			return ErrExpired
 		default:
 			if c > 0 {
-				writer.Write(msg[:c])
+				if err = writer.Write(msg[:c]); err != nil {
+					break process
+				}
 				continue process
 			}
 		}


philips:
LGTM.

philips:
cc @jonboulle

sdjournal/read.go
```diff
@@ -161,7 +161,9 @@ process:
 			return ErrExpired
 		default:
 			if c > 0 {
-				writer.Write(msg[:c])
+				if _, err := writer.Write(msg[:c]); err != nil {
```
alban: With `:=`, the inner "err" variable masks the outer "err" variable and the function will not return any error. Is it intended and/or correct?

yifan-gu: good catch

alban:
LGTM now

flyingmutant:
FYI, this PR breaks the build with

```
./read.go:164: multiple-value writer.Write() in single-value context
```

It looks like the CI does not cover sdjournal, unfortunately.

jonboulle:
Yeah, just noticed that :-(

yifan-gu:
oops, my bad

alban:
Fixed via #175 / #177

PR #177, opened 2016-06-17
sdjournal: fix Writer call

### LINKED ISSUE #175: `go get github.com/coreos/go-systemd/sdjournal` fails
We're using [journald-cloudwatch-logs](https://github.com/saymedia/journald-cloudwatch-logs), which is a Go application that sends our servers' Journald logs to Amazon Web Services' CloudWatch Logs service.

Up until very recently – perhaps the last couple of days – I was happily installing journald-cloudwatch-logs via `go get` as part of a Chef run. However, it's started bombing, and it seems to be failing while installing prerequisites, one of which is go-systemd-sdjournal. I tracked down the failure and was able to reproduce it while installing go-systemd-sdjournal on its own. Here's how the failure looks:

```
[root@server ~]# go get github.com/coreos/go-systemd/sdjournal
# github.com/coreos/go-systemd/sdjournal
/usr/local/src/go/src/github.com/coreos/go-systemd/sdjournal/read.go:164: multiple-value writer.Write() in single-value context
```

The system's running Red Hat Enterprise Linux 7.2.

I don't have any knowledge of Go and largely am just a monkey installing packages, so I'm not comfortable debugging this. I thought I'd bring this up here to see if anyone has any idea what might be going wrong, or if anyone else is experiencing the same problem.

I'm happy to provide any further details. Just let me know what you need.

Thanks!

jonboulle:
Introduced in 33ad232a0a876386f9335dfddb21104d1aa4bf88

Fixes #175

Not noticed in CI because travis isn't testing the sdjournal package as it's
not running systemd.

### DIFF
--- a/sdjournal/read.go
+++ b/sdjournal/read.go
@@ -161,7 +161,7 @@ process:
 			return ErrExpired
 		default:
 			if c > 0 {
-				if err = writer.Write(msg[:c]); err != nil {
+				if _, err = writer.Write(msg[:c]); err != nil {
 					break process
 				}
 				continue process


jonboulle:
Merging as this is a break-fix

PR #179, opened 2016-06-20
travis: run whole testsuite in containers

lucab:
Rework configuration to run testsuite into containerized
environments. This allows to run all tests on travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,27 @@
-language: go
-go: 1.4
+sudo: required
+
+services:
+  - docker
+
+env:
+  global:
+    - GOPATH=/opt
+    - BUILD_DIR=/opt/src/github.com/coreos/go-systemd
+  matrix:
+    - DOCKER_BASE=ubuntu:16.04
+    - DOCKER_BASE=debian:stretch
+
+before_install:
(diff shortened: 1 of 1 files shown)


lucab:
Regarding the "creative" part of this PR, I think I'm quite bending travis here by ditching all of its language framework and going for vanilla containers here. For reviewers, some points to consider:

Cons:
 * testing multiple go-versions will be harder (sticking to packaged golang for simplicity now)
 * increased testing-time/complexity (due to having to setup a full container, run systemd on init and then attach to run the tests)

Pros:
 * allow full testing of all packages (before we were missing those depending on system session/bus)
 * allow testing multiple release / systemd packages (debian testing and latest ubuntu LTS for the moment, but more are easily added)

IMHO pros outweights cons, even if I would consider this kind of setup unconventional (ie. may break if travis decides to ditch it).

jonboulle:
I agree that pros > cons for now and would also consider this something of
an interim solution (it's definitely a step forward from the current lack
of test coverage, which is becoming more critical as we leverage go-systemd
more extensively).

s-urbaniak:
agreed with what has been said above, lgtm

PR #178, opened 2016-06-20
dbus: fix testsuite

lucab:
Some assumptions do not hold anymore on recent debian-derivatives.
Tests updated to avoid failures in modern dbus/systemd environments.

Fixes #165

PR #181, opened 2016-06-20
sdjournal: fix a race in GetEntry test

### LINKED ISSUE #180: sdjournal: possible race in Match/Send/Wait/Next flow
https://github.com/coreos/go-systemd/pull/164#issuecomment-224311422 seems to have highlighted a race between Match/Send/Wait/Next, where `Wait` returns after seeing the matching entry but `Next` reports and EOF. Increasing the Wait timeout doesn't seem to help, so there may be a race somewhere.

After #178 and #179, this is the only remaining testsuite failure. It doesn't happen always, but seem very frequent on debian/ubuntu systems (I've seen it often locally and on containerized travis).

lucab:
Hard-sleep instead of waiting for a journal event.
This fixes a race due to waiting for any events but enumerating only
matching ones.

Closes #180

### DIFF
--- a/sdjournal/journal_test.go
+++ b/sdjournal/journal_test.go
@@ -204,10 +204,7 @@ func TestJournalGetEntry(t *testing.T) {
 		t.Fatalf("Error writing to journal: %s", err)
 	}
 
-	r := j.Wait(time.Duration(1) * time.Second)
-	if r < 0 {
-		t.Fatalf("Error waiting to journal")
-	}
+	time.Sleep(time.Duration(1) * time.Second)
 
 	n, err := j.Next()
 	if err != nil {


lucab:
All together #178, #179, and #180 let [the whole testsuite run on travis without failures](https://travis-ci.org/lucab/go-systemd/builds/138868717).

jonboulle:
@lucab how about just putting them up in a single PR?

lucab:
@jonboulle #179 is some "creative" travis usage which I'm not sure if ok. The other two could have been merged yes, but this last one was unplanned at first as I was expecting #180 to be harder to trace down.

jonboulle:
@lucab is that like "creative accounting"? ;-). 
What are your concerns with that approach exactly? Longer term, we could consider migrating this to our Jenkins infrastructure for more flexibility with different systemds.

lucab:
@jonboulle exactly :smile: 

I expanded the discussion [in that PR](https://github.com/coreos/go-systemd/pull/179#issuecomment-227780746).

I'm not merging the other two testsuite-fixing PRs now, but I'll avoid wasting resources next time. Relevant to this specific PR, there is a thread ongoing on systemd ML: https://lists.freedesktop.org/archives/systemd-devel/2016-June/036946.html

sdjournal/journal_test.go
```diff
@@ -204,10 +204,7 @@ func TestJournalGetEntry(t *testing.T) {
 		t.Fatalf("Error writing to journal: %s", err)
 	}
 
-	r := j.Wait(time.Duration(1) * time.Second)
-	if r < 0 {
-		t.Fatalf("Error waiting to journal")
-	}
+	time.Sleep(time.Duration(1) * time.Second)
```
s-urbaniak: Can you expand on why a hard sleep is better than waiting for the journal? This seems counter-intuitive to me as `time.Sleep` cries for flakes later on.

lucab: From ML thread:
```
sd_journal_wait() will trigger on *any* events, while sd_journal_get_data() 
will apply the filter and find no matching entries.
```

In our context, Wait() will return as soon as there is a new message available in the log even if it doesn't match the filter, in which case Next() will return EOF (if expected entry is not yet available).

This sleep *could* be replaced with something like a Wait&Next with a retrial counter, but it will *not* eliminate eliminate the flakiness of the test: it will then depend on the magic number of retrials and additionally also on how many events happen before our expected entry.

s-urbaniak: We could loop over `sd_journal_wait()`/`sd_journal_get_data()` until we eventually get our message matching the field, but I agree this could end up in an endless loop.

Not really happy about the `time.Sleep` at all, so let's observe this in subsequent CI builds.

lucab: Me neither, but I think sdjournal would need to grow a C API for filtered event-triggering to properly address this usecase. May I go on and merge this as-is for now?

Issue #184, opened 2016-06-22
SdNotify should return an error compliant with with sd_notify()


alepuccetti:
I think it will be better if this function returns 0 in the case that $NOTIFY_SOCKET is not set as stated in [sd_notify()](https://www.freedesktop.org/software/systemd/man/sd_notify.html#Return%20Value) documentation.

PR #185, opened 2016-06-22
daemon: improve SdNotify errors signaling

alepuccetti:
This patch makes SdNotify return value compliant to the sd_notify()
documentation. See sd_notify(3) man page.

Fixes: https://github.com/coreos/go-systemd/issues/184

### DIFF
--- a/daemon/sdnotify.go
+++ b/daemon/sdnotify.go
@@ -2,30 +2,37 @@
 package daemon
 
 import (
-	"errors"
 	"net"
 	"os"
 )
 
-var SdNotifyNoSocket = errors.New("No socket")
-
 // SdNotify sends a message to the init daemon. It is common to ignore the error.
-func SdNotify(state string) error {
+// It returns one of the following:
+// (false, nil) - notification not supported (i.e. NOTIFY_SOCKET is unset)
+// (false, err) - notification supported, but failure happened (e.g. error connecting to NOTIFY_SOCKET or while sending data)
+// (true, nil) - notification supported, data has been sent
+func SdNotify(state string) (sent bool, err error) {
 	socketAddr := &net.UnixAddr{
 		Name: os.Getenv("NOTIFY_SOCKET"),
 		Net:  "unixgram",
 	}
 
+	// NOTIFY_SOCKET not set
 	if socketAddr.Name == "" {
-		return SdNotifyNoSocket
+		return false, nil
 	}
 
 	conn, err := net.DialUnix(socketAddr.Net, nil, socketAddr)
+	// Error connecting to NOTIFY_SOCKET
 	if err != nil {
-		return err
+		return false, err
 	}
 	defer conn.Close()
 
 	_, err = conn.Write([]byte(state))
-	return err
+	// Error sending the message
+	if err != nil {
+		return false, err
+	}
+	return true, nil
 }
--- a/daemon/sdnotify_test.go
+++ b/daemon/sdnotify_test.go
@@ -1,0 +1,75 @@
+// Copyright 2016 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package daemon
+
+import (
+	"io/ioutil"
+	"net"
+	"os"
+	"testing"
+)
+
+// TestSdNotify
+func TestSdNotify(t *testing.T) {
+	notificationSupportedDataSent := "Notification supported, data sent"
+	notificationSupportedFailure := "Notification supported, but failure happened"
+	notificationNotSupported := "Notification not supported"
+
+	testDir, e := ioutil.TempDir("/tmp/", "test-")
+	if e != nil {
+		panic(e)
+	}
+	defer os.RemoveAll(testDir)
+
+	notifySocket := testDir + "/notify-socket.sock"
+	laddr := net.UnixAddr{
+		Name: notifySocket,
+		Net:  "unixgram",
+	}
+	_, e = net.ListenUnixgram("unixgram", &laddr)
+	if e != nil {
+		panic(e)
+	}
+
+	// (true, nil) - notification supported, data has been sent
+	e = os.Setenv("NOTIFY_SOCKET", notifySocket)
+	if e != nil {
+		panic(e)
+	}
+	sent, err := SdNotify(notificationSupportedDataSent)
+	if !sent || err != nil {
+		t.Errorf("TEST: %s FAILED", notificationSupportedDataSent)
+	}
+
+	// (false, err) - notification supported, but failure happened
+	e = os.Setenv("NOTIFY_SOCKET", testDir+"/not-exist.sock")
+	if e != nil {
+		panic(e)
+	}
+	sent, err = SdNotify(notificationSupportedFailure)
+	if sent && err == nil {
+		t.Errorf("TEST: %s FAILED", notificationSupportedFailure)
+	}
+
+	// (false, nil) - notification not supported
+	e = os.Unsetenv("NOTIFY_SOCKET")
+	if e != nil {
+		panic(e)
+	}
+	sent, err = SdNotify(notificationNotSupported)
+	if sent || err != nil {
+		t.Errorf("TEST: %s FAILED", notificationNotSupported)
+	}
+}
--- a/test
+++ b/test
@@ -27,7 +27,7 @@ if [ -z "$GOPATH" ]; then
 	go get -u github.com/coreos/pkg/dlopen
 fi
 
-TESTABLE="activation journal login1 machine1 unit"
+TESTABLE="activation daemon journal login1 machine1 unit"
 FORMATTABLE="$TESTABLE sdjournal dbus"
 if [ -e "/run/systemd/system/" ]; then
 	TESTABLE="${TESTABLE} sdjournal"


alban:
LGTM

jonboulle:
NACK - we should either fully implement the semantics or not at all: sd_notify returns three categories of conditions (negative = error, zero = nothing sent, positive = success), but now we're not distinguishing between the latter two cases:  

> On failure, these calls return a negative errno-style error code. If $NOTIFY_SOCKET was not set and hence no status data could be sent, 0 is returned. If the status was sent, these functions return with a positive return value. In order to support both, init systems that implement this scheme and those which do not, it is generally recommended to ignore the return value of this call.

jonboulle:
How about changing the signature to be (bool, error) => (sent, error)

alban:
@jonboulle Oh, I missed the part where it returns positive values. Your suggested signature sounds good to me.

alepuccetti:
@jonboulle 
> How about changing the signature to return (bool, error) => (sent, error)

What about this:
sd_notify()<0 --> SdNotify return e message "some error here"
sd_notify()==0 --> SdNotify an empty string ""
sd_nodify()>0 --> SdNotify return nil
I think this maps the sd_notify return vaules pretty well

jonboulle:
@alepuccetti I'm not sure what you're proposing - `func SdNotify(state string) string` ? If so, you cannot return `nil` - https://play.golang.org/p/ZJOzL2iOSb

alepuccetti:
@jonboulle I meant that return type `error` with an empty "description" like this
```
var SdNotifyNoSocket = errors.New("")
```
But it might be not very elegant returning an error that is not an error.

jonboulle:
> But it might be not very elegant returning an error that is not an error.
 
correct :-) - I think this is worse than the current situation (which I actually think is fine since it still allows the caller to know whether the message was sent)

alepuccetti:
Ok then, so if someone wants to check this then have to "parse" the error message.

jonboulle:
@alepuccetti No, they compare it against the const that this package exposes! ;-)

alban:
> actually think is fine since it still allows the caller to know whether the message was sent

In practice, applications should only care about errors when the $NOTIFY_SOCKET is set, rather than caring if the message was sent.

I think that requiring programs using go-systemd to compare the error to the constant error that should normally be ignored is a bit clumsy, so I prefer the (sent bool, err error) signature.

It seems that etcd prints a warning even when it is not executed as a systemd unit file:
https://github.com/coreos/etcd/blob/v2.3.7/etcdmain/etcd.go#L178
Is it intended? I get this message when starting `etcd` from the command line instead of a systemd unit file:

```
2016-06-22 17:25:18.677145 E | etcdmain: failed to notify systemd for readiness: No socket
2016-06-22 17:25:18.677184 E | etcdmain: forgot to set Type=notify in systemd service file?
```

/cc @xiang90

xiang90:
@alban Now it only detects systemd is running. Can you send a patch to fix it by also detecting running as a systemd unit?

lucab:
(Let's detach the etcd discussion from here)

@alepuccetti can you please rework this to return `(sent bool, err error)` and also add an initial simple test excercising the "No socket" error codepath?

alepuccetti:
Patch updated

daemon/sdnotify.go
```diff
@@ -2,30 +2,33 @@
...
 	}
 
+	// NOTIFY_SOCKET not set
 	if socketAddr.Name == "" {
-		return SdNotifyNoSocket
+		return false, nil
 	}
 
 	conn, err := net.DialUnix(socketAddr.Net, nil, socketAddr)
+	// Error connecting ot NOTIFY_SOCKET
```
lucab: s/ot/to/

daemon/sdnotify.go
```diff
@@ -2,30 +2,33 @@
...
 
 import (
-	"errors"
 	"net"
 	"os"
 )
 
-var SdNotifyNoSocket = errors.New("No socket")
-
 // SdNotify sends a message to the init daemon. It is common to ignore the error.
```
lucab: docstring should be updated to document the three possible cases:
 * (true, nil) - notification supported, data has been sent
 * (false, nil) - notification not supported
 * (false, err) - notification supported, but failure happened

lucab:
Comments inline. What about a small test covering at least the (false, nil) case?

alepuccetti:
Patch updated. I implemented two out of three tests and sketched the third one.

daemon/sdnotify_test.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"os"
+	"testing"
+)
+
+// TestDaemon
+func TestDaemon(t *testing.T) {
+	// (true, nil) - notification supported, data has been sent
+	// this test should work as long as the test host runs systmed
+	os.Setenv("NOTIFY_SOCKET", "/run/systemd/notify")
```
lucab: This seems a bit brutal. Can we at least check that the file exists as a 0777 socket, and skip this first combination if not?

daemon/sdnotify_test.go
```diff
@@ -0,0 +1,43 @@
...
+	if !sent || err != nil {
+		t.Errorf("TEST: \"notification supported, data sent\" FAILED")
+	}
+	// (false, nil) - notification not supported
+	os.Unsetenv("NOTIFY_SOCKET")
+	sent, err = SdNotify("TEST: notification not supported")
+	if sent || err != nil {
+		t.Errorf("TEST: \"notification not supported\" FAILED")
+	}
+	// TODO (false, err) - notification supported, but failure happened
```
lucab: You should be able to reach this case by pointing $NOTIFY_SOCKET to something invalid like a non-existing file, a path that you cannot write to or that is not a socket.

test
```diff
@@ -27,7 +27,7 @@ if [ -z "$GOPATH" ]; then
 	go get -u github.com/coreos/pkg/dlopen
 fi
 
-TESTABLE="activation journal login1 machine1 unit"
+TESTABLE="activation journal login1 machine1 unit daemon"
 FORMATTABLE="$TESTABLE sdjournal dbus"
 if [ -e "/run/systemd/system/" ]; then
 	TESTABLE="${TESTABLE} sdjournal"
```
lucab: Your test requires a running systemd, so I think it should be placed here instead (guarded by the if) and in $FORMATTABLE too.

daemon/sdnotify_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+package daemon
+
+import (
+	"os"
+	"testing"
+)
+
+// TestDaemon
+func TestDaemon(t *testing.T) {
```
lucab: This should be called TestSdNotify.

lucab:
Added further comments on the testing part.

daemon/sdnotify_test.go
alban: Or implement a sdnotify server that would listen on `/tmp/$(mktemp...)/foo` and use `os.Setenv` on that path instead? In this way, we avoid sending fake events to the real systemd during the test.

lucab: Your proposal is probably better. We don't even need systemd / sdnotify at all, just a temporary bare socketfile to write to.

test
lucab: With @alban suggestion, this is not needed at all.

daemon/sdnotify_test.go
alepuccetti: > we avoid sending fake events to the real systemd during the test.

Because `SdNotify()` has hardcoded `NOTIFY_SOCKET` we need to change its value for the test. This will break the notification system (if exist) for a the time of the test. So we have 2 solutions here:

1) If `NOTIFY_SOCKET` exists we use it and systemd receives the test notifications, I do not think that it is a problem as long as the notifications do not carry a message that can be interpreted as command (e.g. READY=1, STATUS=blablabla). If `NOTIFY_SOCKET` do not exist we create it.

2) We need to change `SdNotify()` in a way that can be specified a different socket `func SdNotify(notifySocket, state string) (sent bool, err error) `. And have these cases:
* notifySocket == nil --> Notification non supported.
* notifySocket == "" --> Use the value in `NOTIFY_SOCKET` to get the socket.
* notifySocket == "VARIABLE_1" --> Use the value in `VARIABLE_1` to get the socket.

I am for the first solution, I do not think that if systemd receives some bogus notifications it is a problem.

alepuccetti:
Patch updated

daemon/sdnotify.go
s-urbaniak: I suggest the following (yet simpler), leave the signature as-is:

`func SdNotify(state string) error`

and return:

return `nil`: notification supported, data has been sent
return `ErrNotSupported`: notification not supported, where `var ErrNotSupported = errors.New("unsupported")`
return `err`: notification supported, but failure happened (`err != ErrNotSupported`)

lucab: @s-urbaniak while the (bool, error) signature is not nice, @alban's concern is that this error is typically ignored (following C practice). You are arguing for the status quo, just removing the "ignore" bits from the docstring. Either is fine with me.

daemon/sdnotify_test.go
```diff
@@ -0,0 +1,64 @@
...
+	"os"
+	"testing"
+	"time"
+)
+
+// TestSdNotify
+func TestSdNotify(t *testing.T) {
+	rand.Seed(time.Now().UnixNano())
+	testDir := fmt.Sprintf("/tmp/test-%d", rand.Int())
+	os.Mkdir(testDir, 0777)
```
lucab: You are looking for https://golang.org/pkg/io/ioutil/#TempDir here.

daemon/sdnotify.go
s-urbaniak: I just read @alban's concern about the clumsiness of my proposal. Well, it may seem clumsy, but that's quite "idiomatic" in Go, the most prominent example being the const `io.EOF`, amongst other possible solutions (see http://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully).

To make it more prominent, we could/should make the behavior very explicit in the godoc.

alban: @s-urbaniak In practice, your `ErrNotSupported` is something that apps should ignore. I prefer to keep the "error" only for real errors: that allows apps to write simpler code:

```
if _, err := SdNotify("READY=1"); err != nil {
  fmt.Printf(err)
  os.Exit(1)
}
```

s-urbaniak: @alban ack, but `io.EOF` is no "real" error either ;-) Either way, I'm fine with the current solution, and am very much willing to avoid to bikeshed about this :bow:

alepuccetti:
This should be it

daemon/sdnotify_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"fmt"
+	"io/ioutil"
+	"net"
+	"os"
+	"testing"
+)
+
+// TestSdNotify
+func TestSdNotify(t *testing.T) {
+	testDir, err := ioutil.TempDir("/tmp/", "test-")
```
lucab: if err is non-nil, we should do a `Fatal()` here.

daemon/sdnotify_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+	notifySocket := testDir + "/notify-socket.sock"
+	laddr := net.UnixAddr{
+		Name: notifySocket,
+		Net:  "unixgram",
+	}
+	_, e := net.ListenUnixgram("unixgram", &laddr)
+	if e != nil {
+		t.Errorf("Error creating the notification socket")
+		panic(e)
```
lucab: Please don't panic in test, a `Fatal()` is enough.

lucab:
Latest nits from side, and then I think it is ready to go. @jonboulle is the new shape of this ok now?

daemon/sdnotify_test.go
jonboulle: panic tends to be used to indicate failure during test set-up that is distinct from the behaviour being tested

daemon/sdnotify_test.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	os.Create(notifySocket)
```
jonboulle: Unhandled error

daemon/sdnotify_test.go
```diff
@@ -0,0 +1,61 @@
...
+	os.Setenv("NOTIFY_SOCKET", notifySocket)
+	fmt.Println(os.Getenv("NOTIFY_SOCKET"))
```
jonboulle: ?

daemon/sdnotify_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+	// (true, nil) - notification supported, data has been sent
+	sent, err := SdNotify("TEST: notification supported, data sent")
```
jonboulle: I'd rather these were in a test table

daemon/sdnotify.go
```diff
@@ -2,30 +2,37 @@
...
-func SdNotify(state string) error {
+// It returns one of the following:
+// (false, nil) - notification not supported
```
jonboulle: (i.e. NOTIFY_SOCKET is unset)

daemon/sdnotify.go
```diff
@@ -2,30 +2,37 @@
...
+// (false, err) - notification supported, but failure happened
```
jonboulle: (e.g. error connecting to NOTIFY_SOCKET or while sending data)

jonboulle:
More nits but generally OK

alepuccetti:
I sill need to do the test table thing, I missed that comment.

alepuccetti:
@jonboulle 
Thinking about the test table, it seems a bit of a stretch. The first two tests are very similar but the last one it unsets the environment variable so writing a for loop to execute each test it is more complex than just write three cases. I think it is ok like this.
If this test was 10 different cases like [rkt_env_test.go](https://github.com/coreos/rkt/blob/master/tests/rkt_env_test.go) I would agree with you on using the table.

test
```diff
@@ -27,7 +27,7 @@ if [ -z "$GOPATH" ]; then
 	go get -u github.com/coreos/pkg/dlopen
 fi
 
-TESTABLE="activation journal login1 machine1 unit"
+TESTABLE="activation journal login1 machine1 unit daemon"
```
jonboulle: Sort please

jonboulle:
@alepuccetti you can probably talk me out of the test table, but there's still various stuff I'd like to see cleaned up in the tests - all of the env setup failure cases should be panics, for example, and there's a fair amount of duplication in string construction etc

alepuccetti:
Patch update, failure seems unrelated

daemon/sdnotify.go
lucab: For reference, bool+error is also how rust-systemd bindings do it: https://github.com/jmesmon/rust-systemd/blob/master/src/daemon.rs#L189

lucab:
Re-triggered, it is now green (I'll keep an eye on `TestConnJobListener` for the future). I'm not sure this addresses all remarks in https://github.com/coreos/go-systemd/pull/185#issuecomment-232965592.

alepuccetti:
Patch updated

daemon/sdnotify_test.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	// (true, nil) - notification supported, data has been sent
+	e = os.Setenv("NOTIFY_SOCKET", notifySocket)
+	if e != nil {
+		panic(e)
+	}
+	sent, err := SdNotify("TEST: notification supported, data sent")
+	if !sent || err != nil {
+		t.Errorf("TEST: \"notification supported, data sent\" FAILED")
```
lucab: I think https://github.com/coreos/go-systemd/pull/185#issuecomment-232965592 meant that this message can be constructed by re-using the test string above it. Same for the other two below.

alepuccetti: The message in `SdNotify()` is not important. I will change it to `""`.

lucab: It is ok to have some content, it make the test easier to read too. Just declare the string beforehand and use it in both places. You can can pass a `%s` format specifier to `Errorf`.

alepuccetti:
This should be it

lucab:
That's it, thanks! LGTM.

PR #186, opened 2016-07-04
sdjournal: fix an implicit seek to head

lucab:
When a JournalReader is initialized with a NumFromTail bigger than
the number of available entries, cursor will be positioned on the
first entry. Upon initial (mandatory) Next invocation, it will
incorrectly start reading from the second entry.

This fixes a pathological case where the first entry of a journal
is always ignored when using NumFromTail.


lucab:
Ref https://github.com/coreos/rkt/issues/2776

s-urbaniak:
per oob: lgtm

jonboulle:
LGTM, nice catch

PR #190, opened 2016-08-12
util: conditionally build CGO functions

### LINKED ISSUE #171: no buildable Go source files
Disclaimer: I'm not normally a Go developer, so some of my terms may not be quite right.

I'm trying to build a Docker plugin that use the `go-plugins-helpers` from Docker and I wanted the binary to run in Docker `alpine`, but `alpine` doesn't have the linked dynamic binaries or whatever, so I was reading up that I need to set `CGO_ENABLED=0` when building. I've tried a few different build commands, including:

```
CGO_ENABLED=0 go build -a -installsuffix cgo -ldflags '-s' -o bin/linux/amd64/local-persist -v
CGO_ENABLED=0 go build -a -tags netgo -a -o bin/linux/amd64/local-persist -v
```

(Tried that last one with and without `CGO_ENABLED` flag. No error, but still won't run in `alpine`.)

I keep getting this error:

```
../github.com/docker/go-plugins-helpers/sdk/unix_listener.go:12:2: no buildable Go source files in /go/src/github.com/coreos/go-systemd/util
make: *** [binary-linux-amd64] Error 1
```

Steps to reproduce yourself:

```
git clone https://github.com/CWSpear/local-persist
cd local-persist
docker run -it -v `pwd`:/go/src/app  golang:onbuild bash
# now you'll be inside the docker container, next run:
export CGO_ENABLED=0
go get -v -d
go build [whatever flags] .
```

See https://github.com/docker/go-plugins-helpers/issues/50

heyitsanthony:
Some projects don't need the CGO functions in the util package;
split out the functions into a separate file so CGO_ENABLED=0
works if those functions are never used.

Fixes #171

### DIFF
--- a/util/util.go
+++ b/util/util.go
@@ -18,96 +18,23 @@
 // than linking against them.
 package util
 
-// #include <stdlib.h>
-// #include <sys/types.h>
-// #include <unistd.h>
-//
-// int
-// my_sd_pid_get_owner_uid(void *f, pid_t pid, uid_t *uid)
-// {
-//   int (*sd_pid_get_owner_uid)(pid_t, uid_t *);
-//
-//   sd_pid_get_owner_uid = (int (*)(pid_t, uid_t *))f;
-//   return sd_pid_get_owner_uid(pid, uid);
-// }
-//
-// int
-// my_sd_pid_get_unit(void *f, pid_t pid, char **unit)
-// {
-//   int (*sd_pid_get_unit)(pid_t, char **);
-//
-//   sd_pid_get_unit = (int (*)(pid_t, char **))f;
-//   return sd_pid_get_unit(pid, unit);
-// }
-//
-// int
-// my_sd_pid_get_slice(void *f, pid_t pid, char **slice)
-// {
-//   int (*sd_pid_get_slice)(pid_t, char **);
-//
-//   sd_pid_get_slice = (int (*)(pid_t, char **))f;
-//   return sd_pid_get_slice(pid, slice);
-// }
-//
-// int
-// am_session_leader()
-// {
-//   return (getsid(0) == getpid());
-// }
-import "C"
 import (
 	"fmt"
 	"io/ioutil"
 	"os"
 	"strings"
-	"syscall"
-	"unsafe"
-
-	"github.com/coreos/pkg/dlopen"
 )
 
-var libsystemdNames = []string{
-	// systemd < 209
-	"libsystemd-login.so.0",
-	"libsystemd-login.so",
-
-	// systemd >= 209 merged libsystemd-login into libsystemd proper
-	"libsystemd.so.0",
-	"libsystemd.so",
-}
+var (
+	ErrNoCGO = fmt.Errorf("go-systemd built with CGO disabled")
+)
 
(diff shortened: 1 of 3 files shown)


lucab:
This sounds good to me.

I'm just wondering if the non-cgo stub should still implement the functions that are being moved out here, just by erroring out. This would keep the same public interface, regardless of tags.

I'm not sure what is the idiomatic Go way here, @jonboulle?

gyuho:
@lucab @jonboulle Can we get this merged?

go-systemd build in etcd is broken.

And I believe this is idiomatic, based on Go source tree.

```
$ find . -name '*_cgo_*'

./src/crypto/x509/root_cgo_darwin.go
./src/runtime/crash_cgo_test.go
```

And more

```
$ find . -name '*cgo_*'

./src/net/cgo_unix_test.go
./src/net/cgo_bsd.go
...
```

Thanks!

util/util.go
```diff
@@ -18,190 +18,13 @@
...
-	// systemd >= 209 merged libsystemd-login into libsystemd proper
-	"libsystemd.so.0",
-	"libsystemd.so",
-}
-
-// GetRunningSlice attempts to retrieve the name of the systemd slice in which
-// the current process is running.
-// This function is a wrapper around the libsystemd C library; if it cannot be
-// opened, an error is returned.
-func GetRunningSlice() (slice string, err error) {
```
lucab: Function declaration and docstring should probably be retained, otherwise this part will be missing from godoc and we would get different public symbols depending on build tags. It can be simply implemented as a stub that always errors out. Same for the other functions moved out from this file.

lucab:
I had another look at this, and I think that the current approach may be non-optimal: some public symbols will depend on build tags and will disappear from godoc (I guess, not sure).

Thus, I'd suggest a different way this could be done, by splitting current `util.go` into three parts:
 * `util.go` without any tag, only with the pure-Go functions (`GetMachineID` and `IsRunningSystemd`)
 * `util_cgo.go` gated behind a cgo tag, as you already did
 * `util_stub.go` gated behind a not-cgo tag, with stubs and docstring to match the cgo counterparts

heyitsanthony:
@lucab all fixed up; I made the public functions call into private build-specific functions. It seemed weird to have the full godoc in `_stub.go` when the code there does nothing. Thanks!

util/util_stub.go
```diff
@@ -0,0 +1,29 @@
...
+
+// +build !cgo
+
+package util
+
+import (
+	"fmt"
+)
+
+var errNoCGO = fmt.Errorf("go-systemd built with CGO disabled")
```
lucab: This should be moved to `util.go` in order to let consumers error-match on this in all situations.

lucab:
> I made the public functions call into private build-specific functions. It seemed weird to have the full godoc in _stub.go when the code there does nothing.

I agree this is even better than my half-botched proposal, thanks! Just one final remark inline, other than that it should be ok.

heyitsanthony:
@lucab ok, ErrNoCGO exported

lucab:
Great, LGTM!

PR #191, opened 2016-08-15
Don't specify NextProtos so that we can take advantage of HTTP/2

aaronjwood:
If we don't specify `NextProtos` HTTP/2 support will be enable when possible. See https://github.com/golang/go/issues/14374#issuecomment-187939699 and https://github.com/golang/go/commit/b5f0aff49503e31002b33198e06708e263c445a7

### DIFF
--- a/activation/listeners.go
+++ b/activation/listeners.go
@@ -48,8 +48,6 @@ func TLSListeners(unsetEnv bool, tlsConfig *tls.Config) ([]net.Listener, error) 
 	}
 
 	if tlsConfig != nil && err == nil {
-		tlsConfig.NextProtos = []string{"http/1.1"}
-
 		for i, l := range listeners {
 			// Activate TLS only for TCP sockets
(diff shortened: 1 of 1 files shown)


aaronjwood:
Actually, I don't see this `TLSListeners` method getting called anywhere. Should it be removed?

aaronjwood:
Going to close this as it looks like `TLSListeners` is dead code :)

jonboulle:
This isn't really dead code - it's an exported function that we happen to not use anywhere else internally in this repository ;-). I think this is still appropriate so feel free to re-open

aaronjwood:
Okay, reopening then :)

jonboulle:
ok to test

jonboulle:
LGTM - any chance you can tweak your commit message though?
`activation: remove setting of NextProtos from tls config` or similar

aaronjwood:
Sure, just altered it.

PR #192, opened 2016-08-15
util: fix import path

gyuho:
@heyitsanthony @lucab


heyitsanthony:
oops, lgtm

lucab:
Looks like utils is not covered by any tests at all. Mmh, https://github.com/coreos/go-systemd/issues/193.

PR #194, opened 2016-08-23
sdjournal: fix memory leak in cursor handling

amenzhinsky:
https://github.com/systemd/systemd/blob/622a0f628cee51851ad00856f9efddedf0799edb/src/journal/sd-journal.c#L936

Since `asprintf`'s `dst` argument has to be freed

Like it's done here: https://github.com/systemd/systemd/blob/29a753df7682424a0ea505698d548f85c514fad5/src/shared/logs-show.c#L471

### DIFF
--- a/sdjournal/journal.go
+++ b/sdjournal/journal.go
@@ -736,6 +736,8 @@ func (j *Journal) GetEntry() (*JournalEntry, error) {
 	entry.MonotonicTimestamp = uint64(monotonicUsec)
 
 	var c *C.char
+	defer C.free(unsafe.Pointer(c))
+
 	r = C.my_sd_journal_get_cursor(sd_journal_get_cursor, j.cjournal, &c)
 	if r < 0 {
 		return nil, fmt.Errorf("failed to get cursor: %d", syscall.Errno(-r))
@@ -834,12 +836,13 @@ func (j *Journal) GetMonotonicUsec() (uint64, error) {
 
 // GetCursor gets the cursor of the current journal entry.
 func (j *Journal) GetCursor() (string, error) {
-	var d *C.char
-
 	sd_journal_get_cursor, err := j.getFunction("sd_journal_get_cursor")
 	if err != nil {
 		return "", err
 	}
+
+	var d *C.char
+	defer C.free(unsafe.Pointer(d))
 
 	j.mu.Lock()
 	r := C.my_sd_journal_get_cursor(sd_journal_get_cursor, j.cjournal, &d)


lucab:
ok to test

lucab:
Thanks @amenzhinsky. You are right, we are not properly cleaning up at the FFI border. It looks like there is another place where we are missing it, here: https://github.com/coreos/go-systemd/blob/master/sdjournal/journal.go#L837

Can you please:
 * adjust the commit title as recommended [in the docs](https://github.com/coreos/go-systemd/blob/master/CONTRIBUTING.md#format-of-the-commit-message)
 * plug the other memory leak above too

amenzhinsky:
@lucab, done, please have a look.

It's like `GetEntry` still has some more leaking problems, I'll spend some more time figuring it out today.

sdjournal/journal.go
```diff
@@ -835,6 +837,7 @@ func (j *Journal) GetMonotonicUsec() (uint64, error) {
 // GetCursor gets the cursor of the current journal entry.
 func (j *Journal) GetCursor() (string, error) {
 	var d *C.char
+	defer C.free(unsafe.Pointer(d))
```
lucab: This (together with the var declaration perhaps?) should be moved past the `getFunction` call, or it may be freeing a pointer that has never gone through a malloc.

amenzhinsky: Moved it down below the `getFunction` call

lucab:
LGTM. I'm taking this for the moment, and we can address other leaks separately.

> It's like GetEntry still has some more leaking problems, I'll spend some more time figuring it out today.

I had a very quick look at it right now and didn't spot any obvious ones. Do you have a small reproducer where you observed this?

amenzhinsky:
@lucab, here's an example, if run it for a while you'll see that memory usage is constantly growing

```go
package main

import (
	"fmt"
	"runtime"
)

func main() {

	if err != nil {
		panic(err)
	}

	if _, err := j.Next(); err != nil {
		panic(err)
	}

	for i := 0;; i++ {
		_, err := j.GetEntry()

		if err != nil {
			panic(err)
		}

		fmt.Println(i)

		if i % 1000 == 0 {
			runtime.GC()
		}
	}
}
```

lucab:
Well, I also see the memleak, and valgrind trace it down to the string originating in `sd_journal_get_cursor`.

The weird part is that turning the deferred-free we added here into a simple free right after we are done with assigning `entry.Cursor` seems to plug the leak. But I'm not completely sure what's going on. @jonboulle any ideas perhaps?

amenzhinsky:
I can confirm that `sd_journal_get_cursor` causes the memleak, I guess that's systemd's bug per se

jonboulle:
See #195

amenzhinsky:
@jonboulle, yes that works!

PR #195, opened 2016-08-23
sdjournal: free correct pointer

jonboulle:
Since defer creates a closure, it ends up freeing the wrong
pointer in these situations (i.e. not the one that ends up
getting allocated by asprintf during the actual sdjournal
call).


crawford:
-_- LGTM

jonboulle:
tweaked slightly in case something goes awry in the journald call

crawford:
Smart

jonboulle:
@lucab ?

lucab:
@jonboulle let's take the second commit. Also, thanks for gazing into the abyss here :beer:.

amenzhinsky:
I checked it and it works, nice job :+1:

PR #197, opened 2016-08-24
sdjournal: fix concurent map access

amenzhinsky:
libsystemdFunctions is shared between Journal instances that can cause
concurrency issues when multiple objects modify it at the same time

```
fatal error: concurrent map writes

goroutine 10 [running]:
runtime.throw(0x4b9623, 0x15)
	/usr/lib/go/src/runtime/panic.go:566 +0x95 fp=0xc420028598 sp=0xc420028578
runtime.mapassign1(0x4a0a80, 0xc420014300, 0xc4200286c8, 0xc4200286c0)
	/usr/lib/go/src/runtime/hashmap.go:458 +0x8ef fp=0xc420028680 sp=0xc420028598
github.com/coreos/go-systemd/sdjournal.(*Journal).getFunction(0xc4200fe020, 0x4b863e, 0xf, 0x0, 0x0, 0x0)
````

Code to reproduce the issue:
```go
package main

import (
	"sync"
)

func main() {
	wg := sync.WaitGroup{}
	wg.Add(10)

	for i := 0; i < 10; i++ {
		go func() {

			if err != nil {
				panic(err)
			}

			if _, err = j.Next(); err != nil {
				panic(err)
			}

			for k := 0; k < 1000; k++ {
				if _, err = j.GetEntry(); err != nil {
					panic(err)
				}

			}

			wg.Done()
		}()
	}

	wg.Wait()
}
```

I decided to apply singleton pattern for the libsystemd functions and move then to a separate file.
What do you think?

### DIFF
--- a/sdjournal/functions.go
+++ b/sdjournal/functions.go
@@ -1,0 +1,66 @@
+// Copyright 2015 RedHat, Inc.
+// Copyright 2015 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sdjournal
+
+import (
+	"github.com/coreos/pkg/dlopen"
+	"sync"
+	"unsafe"
+)
+
+var (
+	// lazy initialized
+	libsystemdHandle *dlopen.LibHandle
+
+	libsystemdMutex     = &sync.Mutex{}
+	libsystemdFunctions = map[string]unsafe.Pointer{}
+	libsystemdNames     = []string{
+		// systemd < 209
+		"libsystemd-journal.so.0",
+		"libsystemd-journal.so",
+
+		// systemd >= 209 merged libsystemd-journal into libsystemd proper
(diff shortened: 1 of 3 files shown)


vcaputo:
Makes sense, libdl and its associated state is global for the process, it was inappropriate to serialize per-Journal, lgtm.

sdjournal/functions.go
```diff
@@ -0,0 +1,71 @@
+package sdjournal
```
jonboulle: Please add a copyright header

jonboulle:
Nice catch.

sdjournal/functions.go
```diff
@@ -0,0 +1,71 @@
...
+			return nil, err
+		}
+
+		libsystemdFunctions[name] = f
+	}
+
+	return f, nil
+}
+
+func getHandle() (*dlopen.LibHandle, error) {
```
lucab: This function (with the deferred cleaner) looks too complex for what it is doing, as it retains some logic from its previous place. I think the `dlopen.GetHandle` can be simply done inside the `getFunction` above.

lucab:
This looks like another iteration of https://github.com/coreos/go-systemd/pull/169#issuecomment-224296087 and https://github.com/coreos/go-systemd/pull/170.

I also agree this is a smarter way to achieve it. I've left some comments inline, as I think we can get rid of some more legacy logic there.

sdjournal/functions.go
amenzhinsky: done

sdjournal/functions.go
amenzhinsky: Removed the deffered at all, since it useless now

lucab:
This LGTM now, but I'm leaving some time to @jonboulle for a final pass if he likes.

jonboulle:
LGTM

PR #198, opened 2016-09-06
Syntax highlight activation readme

eklitzke:
Two really minor changes:
 * add syntax highlighting for the bash example in the activation readme
 * fix a really pedantic quoting issue when setting `GOPATH`


jonboulle:
LGTM, thanks

PR #202, opened 2016-10-12
sdjournal: add GetUniqueValues

darkonie:
add functionality to get all unique values for a given field, equivalent of 
`journalctl -F <field_name>`.
https://www.freedesktop.org/software/systemd/man/sd_journal_query_unique.html

### DIFF
--- a/sdjournal/journal.go
+++ b/sdjournal/journal.go
@@ -255,6 +255,33 @@ package sdjournal
 //   return sd_journal_enumerate_data(j, data, length);
 // }
 //
+// int
+// my_sd_journal_query_unique(void *f, sd_journal *j, const char *field)
+// {
+//   int(*sd_journal_query_unique)(sd_journal *, const char *);
+//
+//   sd_journal_query_unique = f;
+//   return sd_journal_query_unique(j, field);
+// }
+//
+// int
+// my_sd_journal_enumerate_unique(void *f, sd_journal *j, const void **data, size_t *length)
+// {
+//   int(*sd_journal_enumerate_unique)(sd_journal *, const void **, size_t *);
+//
+//   sd_journal_enumerate_unique = f;
+//   return sd_journal_enumerate_unique(j, data, length);
+// }
+//
+// void
+// my_sd_journal_restart_unique(void *f, sd_journal *j)
+// {
+//   void(*sd_journal_restart_unique)(sd_journal *);
+//
+//   sd_journal_restart_unique = f;
+//   sd_journal_restart_unique(j);
+// }
+//
 import "C"
 import (
 	"bytes"
@@ -937,4 +964,61 @@ func (j *Journal) GetUsage() (uint64, error) {
 	}
 
 	return uint64(out), nil
+}
+
+// GetUniqueValues returns all unique values for a given field.
+func (j *Journal) GetUniqueValues(field string) ([]string, error) {
+	var result []string
+
+	sd_journal_query_unique, err := getFunction("sd_journal_query_unique")
+	if err != nil {
+		return nil, err
+	}
+
+	sd_journal_enumerate_unique, err := getFunction("sd_journal_enumerate_unique")
+	if err != nil {
+		return nil, err
+	}
+
+	sd_journal_restart_unique, err := getFunction("sd_journal_restart_unique")
+	if err != nil {
+		return nil, err
+	}
+
+	j.mu.Lock()
+	defer j.mu.Unlock()
+
+	f := C.CString(field)
+	defer C.free(unsafe.Pointer(f))
+
+	r := C.my_sd_journal_query_unique(sd_journal_query_unique, j.cjournal, f)
+
+	if r < 0 {
+		return nil, fmt.Errorf("failed to query journal: %d", syscall.Errno(-r))
+	}
+
+	// Implements the SD_JOURNAL_FOREACH_UNIQUE macro from sd-journal.h
+	var d unsafe.Pointer
+	var l C.size_t
+	C.my_sd_journal_restart_unique(sd_journal_restart_unique, j.cjournal)
+	for {
+		r = C.my_sd_journal_enumerate_unique(sd_journal_enumerate_unique, j.cjournal, &d, &l)
+		if r == 0 {
+			break
+		}
+
+		if r < 0 {
+			return nil, fmt.Errorf("failed to read message field: %d", syscall.Errno(-r))
+		}
+
+		msg := C.GoStringN((*C.char)(d), C.int(l))
+		kv := strings.SplitN(msg, "=", 2)
+		if len(kv) < 2 {
+			return nil, fmt.Errorf("failed to parse field")
+		}
+
+		result = append(result, kv[1])
+	}
+
+	return result, nil
 }
--- a/sdjournal/journal_test.go
+++ b/sdjournal/journal_test.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math/rand"
 	"os"
 	"strings"
 	"testing"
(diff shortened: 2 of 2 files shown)


jonboulle:
OK to test

rktbot <notifications@github.com> schrieb am Mi., 12. Okt. 2016, 23:20:

darkonie:
@jonboulle would you be able to have a look at this PR?

sdjournal/journal_test.go
```diff
@@ -343,3 +343,44 @@ func TestJournalReaderSmallReadBuffer(t *testing.T) {
...
+
+	uniqueString := fmt.Sprintf("JOURNALTEST-%d", time.Now().UnixNano())
+	testEntries := []string{"A", "B", "C", "D"}
+	for _, v := range testEntries {
+		if err := journal.Send(uniqueString, journal.PriInfo, map[string]string{"JOURNALTEST": v}); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	time.Sleep(time.Second)
```
jonboulle: How can we avoid this?

sdjournal/journal_test.go
```diff
@@ -343,3 +343,44 @@ func TestJournalReaderSmallReadBuffer(t *testing.T) {
...
+	values, err := j.GetUniqueValues("JOURNALTEST")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(values) < 4 {
+		t.Fatalf("Expect values length >= 4. Got %d", len(values))
+	}
+
+	contains := func(s []string, v string) bool {
```
jonboulle: Rather define this at top level outside this function

sdjournal/journal.go
```diff
@@ -938,3 +965,60 @@ func (j *Journal) GetUsage() (uint64, error) {
...
+	var l C.size_t
+	C.my_sd_journal_restart_unique(sd_journal_restart_unique, j.cjournal)
+	for {
+		r = C.my_sd_journal_enumerate_unique(sd_journal_enumerate_unique, j.cjournal, &d, &l)
+		if r == 0 {
+			break
+		}
+
+		if r < 0 {
+			return result, fmt.Errorf("failed to read message field: %d", syscall.Errno(-r))
```
jonboulle: Any particular reason to return a partial result rather than nil in error cases?

sdjournal/journal_test.go
lucab: There are no filters involved here, so I think it should be possible to wait. Perhaps a wait after each send to be sure that all four values are flushed to journal?

sdjournal/journal_test.go
```diff
@@ -343,3 +343,44 @@ func TestJournalReaderSmallReadBuffer(t *testing.T) {
...
+	testEntries := []string{"A", "B", "C", "D"}
+	for _, v := range testEntries {
+		if err := journal.Send(uniqueString, journal.PriInfo, map[string]string{"JOURNALTEST": v}); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	time.Sleep(time.Second)
+
+	values, err := j.GetUniqueValues("JOURNALTEST")
```
lucab: I'd prefer for the key to be something unique per test run, in order to be sure previous tests are not interfering.

sdjournal/journal_test.go
darkonie: done

sdjournal/journal.go
darkonie: fixed

sdjournal/journal_test.go
darkonie: done

sdjournal/journal_test.go
darkonie: done

darkonie:
@lucab could you have another round of review. Addressed all comments from the previous review.
thank you

darkonie:
@lucab @jonboulle could someone take a look. Thanks

sdjournal/journal_test.go
```diff
@@ -343,3 +344,54 @@ func TestJournalReaderSmallReadBuffer(t *testing.T) {
...
+
+func TestJournalGetUniqueValues(t *testing.T) {
+	j, err := NewJournal()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	defer j.Close()
+
+	uniqueString := generateRandomField(20)
```
lucab: Any reason for this custom generator instead of a simpler current-unix-epoch?

sdjournal/journal_test.go
```diff
@@ -343,3 +344,54 @@ func TestJournalReaderSmallReadBuffer(t *testing.T) {
...
+	testEntries := []string{"A", "B", "C", "D"}
+	for _, v := range testEntries {
+		err := journal.Send("TEST: "+uniqueString, journal.PriInfo, map[string]string{uniqueString: v})
+		if err != nil {
+			t.Fatal(err)
+		}
+		time.Sleep(time.Millisecond * 250)
```
lucab: My previous suggestion was actually to have something similar to [waitAndNext]( https://github.com/coreos/go-systemd/blob/master/sdjournal/journal_test.go#L110) here, so that you can be confident that your latest entry is available when reading. Does it sound reasonable? Have you experienced races here before introducing the sleep?

sdjournal/journal_test.go
lucab: I see, I got confused as one of the tests above does it, but for the value. Nevermind, let's keep it as is.

sdjournal/journal_test.go
lucab: That one has a match filter, which doesn't play properly with wait (see https://github.com/coreos/go-systemd/pull/181).

darkonie: unfortunately `waitAndNext` is racy when we send a several messages in a loop. Sometimes the test passes and sometime does not. I pushed the implementation that uses `golang.org/x/net/context` since you guys are using go 1.6 for testing but it seems to fail. Starting with go 1.7 `context` is a standard library package. https://golang.org/doc/go1.7#context

Do you want to adjust your tests and add `golang.org/x/net/context` for go 1.6 ? Or do you want me to implement it in a different way?

lucab: > unfortunately waitAndNext is racy when we send several messages in a loop. Sometimes the test passes and sometime does not.

This is bad news, the wait is probably getting awaken by another event before your entries are hitting the journal.

I see you are basically implementing a busy loop around `GetUniqueValues()`, which is still a crude hack around missing a proper wait message. I don't want to cause more pain on this PR, so let's just go back to a plain sleep for now and we'll follow-up with a proper `waitOnMatch()` here. Can you please revert to sleeping like in the initial iteration of this PR, and leave there a TODO for the missing helper? And sorry again for the back and forth.

darkonie: thanks you @lucab , PR has been updated.

lucab:
LGTM, we'll handle the TODO separately. @jonboulle?

sdjournal/journal.go
```diff
@@ -938,3 +965,60 @@ func (j *Journal) GetUsage() (uint64, error) {
...
 	return uint64(out), nil
 }
+
+// GetUniqueValues returns all unique values for a given field.
+func (j *Journal) GetUniqueValues(field string) ([]string, error) {
+	var result []string
+
+	sd_journal_query_unique, err := getFunction("sd_journal_query_unique")
+	if err != nil {
+		return result, err
```
jonboulle: Sorry to nit you, but error handling style again: it's preferable to explicitly return `nil, err`, rather than e.g. returning this implicitly-empty slice. Rather, move the result slice declaration down to line 1001-1002 just before it is used. (This also fixes the consistency of returning `result, err` up here but `nil, err` within the for loop below...)

sdjournal/journal.go
```diff
@@ -938,3 +965,60 @@ func (j *Journal) GetUsage() (uint64, error) {
...
+	if r < 0 {
+		return result, fmt.Errorf("failed to query journal: %d", syscall.Errno(-r))
+	}
+
+	// Implements the SD_JOURNAL_FOREACH_UNIQUE macro from sd-journal.h
+	var d unsafe.Pointer
+	var l C.size_t
+	C.my_sd_journal_restart_unique(sd_journal_restart_unique, j.cjournal)
+	for {
+		r = C.my_sd_journal_enumerate_unique(sd_journal_enumerate_unique, j.cjournal, &d, &l)
```
jonboulle: Personal preference would be to codify the assignment and conditional into the loop, i.e.

```
for r = C.my_sd_journal_enumerate_unique(....); 
    r != 0 {
...
}
```
but I don't feel too strongly about it.

sdjournal/journal.go
darkonie: yes, this makes sense. Fixed.

sdjournal/journal.go
darkonie: i followed an example from `GetEntry` https://github.com/coreos/go-systemd/blob/master/sdjournal/journal.go#L687
probably better to leave it as is? or do you want to change that in a separate PR?

jonboulle: Fine as is. thanks for your patience!

PR #203, opened 2016-10-14
dbus: add additional tests for dbus methods

dongsupark:
Add additional unit tests for the following DBUS methods.
    
* RestartUnit
* TryRestartUnit
* ReloadUnit
* ReloadOrRestartUnit
* ReloadOrTryRestartUnit
* KillUnit
* ResetFailedUnit
* Reload

### DIFF
--- a/dbus/methods_test.go
+++ b/dbus/methods_test.go
@@ -111,6 +111,9 @@ func TestStartStopUnit(t *testing.T) {
 	}
 
 	units, err := conn.ListUnits()
+	if err != nil {
+		t.Fatal(err)
+	}
 
 	unit := getUnitStatus(units, target)
 
@@ -130,6 +133,9 @@ func TestStartStopUnit(t *testing.T) {
 	<-reschan
 
 	units, err = conn.ListUnits()
+	if err != nil {
+		t.Fatal(err)
+	}
 
 	unit = getUnitStatus(units, target)
 
@@ -452,6 +458,9 @@ func TestStartStopTransientUnit(t *testing.T) {
 	}
 
 	units, err := conn.ListUnits()
+	if err != nil {
+		t.Fatal(err)
+	}
 
 	unit := getUnitStatus(units, target)
 
@@ -471,6 +480,9 @@ func TestStartStopTransientUnit(t *testing.T) {
 	<-reschan
 
 	units, err = conn.ListUnits()
+	if err != nil {
+		t.Fatal(err)
+	}
 
 	unit = getUnitStatus(units, target)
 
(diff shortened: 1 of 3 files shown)


dongsupark:
CI fails only on Debian Stretch, because `ReloadUnit()` doesn't return an expected value. OTOH it works fine on Ubuntu 16.04, as well as on Fedora 24.
Debian Stretch runs with systemd v231, while other distros are with v229. 
Interesting. I'll have a deeper look.

dongsupark:
Updated. PTAL.

Fortunately, the failure I saw earlier today had nothing to with systemd upstream. It was just because `/bin/kill` didn't exist in the Debian Docker image. Fixed by calling `systemctl kill` instead. Besides I also fixed other minor issues to get it working on Ubuntu, Debian, and Fedora.

lucab:
ok to test

dbus/methods_test.go
```diff
@@ -130,14 +135,209 @@ func TestStartStopUnit(t *testing.T) {
...
+		t.Fatalf("Test unit not found in list")
+	} else if unit.ActiveState != "active" {
+		t.Fatalf("Test unit not active")
+	}
+
+	// Reload the unit
+	reschan = make(chan string)
+
+	// NOTE: As ReloadUnit() sometimes returns "failed", let's
+	// retry 3 times until it's returning the correct status.
```
dongsupark: @lucab I've seen the race only on Fedora 24 with systemd v229. OTOH it seems to work fine on Debian/Ubuntu. The race doesn't always happen. Not sure if it's an upstream issue or a go-systemd issue.

dbus/methods_test.go
```diff
@@ -523,6 +732,107 @@ func TestStartStopTransientScope(t *testing.T) {
...
+	maxAttempts := 3
+	for i := 0; i < maxAttempts; i++ {
+		conn.KillUnit(target, int32(syscall.SIGTERM))
+
+		units, err := conn.ListUnits()
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		unit := getUnitStatus(units, target)
```
dongsupark: @lucab Hmm, that sounds like a good idea. Will try it. I also think such a polling is ugly.

dongsupark:
Updated. PTAL.

* In `TestReloadUnit()` and `TestKillUnit()`, replaced arbitrary polling with a pair of unit subscription and waiting for the events. As `dbus.Conn.Subscribe()` registers a signal `PropertiesChanged` to `org.freedesktop.DBus.Properties`, caller only needs to wait for an event channel to check the returned unit status.

* Split a part of `TestReloadUnit` into `TestReloadOrRestartUnit` for better readability.

lucab:
This LGTM, thanks. @jonboulle do you want to have a pass on this?

lucab:
No concerns from @jonboulle, merging.

dongsupark:
@lucab thanks!

lucab:
@dongsupark thank you for taking care of the testsuite!

PR #204, opened 2016-10-15
sdjournal: fix return values type

darkonie:
fix inconsistency in return types for journal functions: `NextSkip`,
`Previous`, `PreviousSkip`. Currently they will return (uint64, error),
however according to `sd-journal.h` it should be (int, error)

Journal.Next returns (int, error)
Journal.Previous returns (uint64, error)

https://github.com/systemd/systemd/blob/master/src/systemd/sd-journal.h#L98

### DIFF
--- a/sdjournal/journal.go
+++ b/sdjournal/journal.go
@@ -474,10 +474,10 @@ func (j *Journal) FlushMatches() {
 }
 
 // Next advances the read pointer into the journal by one entry.
-func (j *Journal) Next() (int, error) {
+func (j *Journal) Next() (uint64, error) {
 	sd_journal_next, err := getFunction("sd_journal_next")
 	if err != nil {
-		return -1, err
+		return 0, err
 	}
 
 	j.mu.Lock()
@@ -485,10 +485,10 @@ func (j *Journal) Next() (int, error) {
 	j.mu.Unlock()
 
 	if r < 0 {
-		return int(r), fmt.Errorf("failed to iterate journal: %d", syscall.Errno(-r))
+		return 0, fmt.Errorf("failed to iterate journal: %d", syscall.Errno(-r))
 	}
 
-	return int(r), nil
+	return uint64(r), nil
 }
 
 // NextSkip advances the read pointer by multiple entries at once,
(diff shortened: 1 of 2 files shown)


darkonie:
@lucab @jonboulle could someone take a look at this PR. Thanks

lucab:
ok to test

lucab:
This seems reasonable to me, at a glance. @darkonie did you bump into issues anywhere due to this or is it just some general cleanup? @jonboulle do you have insight why the initial signature didn't match the C counterpart?

jonboulle:
Probably because this library was intended to be closer to idiomatic Go than a completely transparent wrapper around the C bindings. I don't buy that we need to be faithful to `sd-journal.h`; we're using a language with multiple returns, so we don't have to overload errors into the return value with the "r < 0" kludge.

What is the benefit of changing this?

darkonie:
@lucab i am working on a journal reader and one of the requirement is to read the journal not only from top to bottom, but also from bottom to top. And one thing i noticed was Next() and Previous() return different types (int and uint64).

jonboulle:
Fair point, but I'd suggest we change in the other direction then to make Next consistent with the others.

darkonie:
@jonboulle we can't because https://github.com/coreos/go-systemd/blob/master/sdjournal/journal.go#L480

darkonie:
also this will fail 
https://github.com/coreos/go-systemd/blob/master/sdjournal/journal.go#L544
if we received -1, so that is definetly a bug

jonboulle:
Generally in Go error propagation has a function with return signature `(interface{}, error)` returning either `(non-nil, nil)` or `(nil, non-nil)` - so in this case, `(0, err)` (and never `(-34, err)`). There's no reason to return anything else _unless_ there's semantic value to the actual value - which would be a good reason to accept the change you've proposed.

jonboulle:
That is indeed a bug, good catch. My preference is still to move to `uint64, error` everywhere unless there's need to propagate systemd's return values.

darkonie:
@jonboulle changed Next to return uint64, please review.

jonboulle:
Thanks! LGTM - @lucab ?

lucab:
Also LGTM.

jonboulle:
For posterity, this is likely a _breaking change_ for any consumers of the library using this function, but i) consumers SHOULD vendor the library, and ii) we do versioned releases, and this will necessitate a version bump.

PR #205, opened 2016-10-27
util: add tests for util package

### LINKED ISSUE #193: util module is not being tested
Currently the `util` module is not covered by any tests. We should start adding at least some basic ones, which would help avoiding breaking things (e.g. https://github.com/coreos/go-systemd/pull/192) during refactoring.

hodgesds:
add tests for the util module

Fixes #193

Not sure if all the assertions are correct on the return values, but it does add a bit of coverage.

### DIFF
--- a/util/util_test.go
+++ b/util/util_test.go
@@ -1,0 +1,74 @@
+// Copyright 2016 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package util
+
+import (
+	"testing"
+)
+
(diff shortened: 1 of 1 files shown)


lucab:
ok to test

lucab:
Thanks for the testcases, @hodgesds! Just a couple of comments:
 * can you please add a copyright header on top of it?
 * I think this test should be gated behind a `cgo` buildtag, they are not much useful without it
 * Our test runners are fine with it, but I think those tests should exit gracefully if systemd is missing in the build/test environment. Can we short-circuit and skip when `!IsRunningSystemd`?

hodgesds:
>can you please add a copyright header at the top of the file?
Will do!

>I think these tests should be gated behind a cgo buildtag, they are not much useful without it
Makes sense.

>Our test runners are fine with it, but I think those tests should exit gracefully if systemd is missing in the build/test environment. Can we short-circuit and skip when !IsRunningSystemd?
I wasn't sure on the desired behavior but I like that approach.

Will try to get an updated PR out later tonight, thanks!

util/util_test.go
```diff
@@ -0,0 +1,53 @@
...
+	s, err := CurrentUnitName()
+	if err != nil {
+		t.Error(err.Error())
+	}
+	if s == "" {
+		t.Error("CurrentUnitName returned a empty string")
+	}
+}
+
+func TestIsRunningSystemd(t *testing.T) {
```
jonboulle: Bit confused about the point of this one...

lucab:
Sorry for the delay, LGTM.

Issue #207, opened 2016-11-01
Investigate one-to-one dbus calling


lucab:
go-systemd currently relies on an existing dbus system bus in order to invoke systemd methods. However, this may be not strictly needed as systemd itself exposes a UNIX socket for one-to-one dbus calling at `/var/run/systemd/private`.

It may be possible to implement a fallback mechanism, in order to retry connecting to the p2p endpoint if the system bus doesn't exist. However this needs a bit more investigation to determine if this works in a transparent way, and the same set of methods/properties are available.

lucab:
Assigning to @squeed for the initial investigation. This may help removing a dbus requirement from some tools expected to run in minimal environments, like runc.

squeed:
This is already supported, but needs to be explicitly requested. If we like, we could change the logic of `New()` to reflect systemd's logic: connect to the private bus if `uid == 0`.

jonboulle:
What guarantee do we have that it's 1:1 with dbus? I'm wary because of https://github.com/systemd/systemd/issues/3610#issuecomment-229679807

squeed:
We don't have that guarantee per se. However, all the current systemd tools operate this way - they try kdbus, then private socket, then public bus.
We have some important scenarios where we need to talk to systemd without dbus running, so adding this as a fallback would be really useful.

lucab:
No explicit guarantee, but systemd internally is doing this [kind of automatic fallback](https://github.com/systemd/systemd/blob/v231/src/shared/bus-util.c#L585) so it seems to handle all cases transparently without special-casing endpoints on different buses. I think this gave us some confidence margin that there won't be big differences.

Just to be more conservative, second iteration of #208 switched to using the p2p mode only as a fallback if the original dbus connection fails. Sounds more reasonable?

PR #208, opened 2016-11-01
dbus: make New() fall back to direct systemd connection

squeed:
This matches the logic in systemd - connect to the private socket first,
if running as root. Fall back to the standard system DBus otherwise.

Fixes: #207

### DIFF
--- a/dbus/dbus.go
+++ b/dbus/dbus.go
@@ -83,9 +83,19 @@ type Conn struct {
 	}
 }
 
-// New establishes a connection to the system bus and authenticates.
+// New establishes a connection to any available bus and authenticates.
 // Callers should call Close() when done with the connection.
 func New() (*Conn, error) {
+	conn, err := NewSystemConnection()
+	if err != nil && os.Geteuid() == 0 {
+		return NewSystemdConnection()
+	}
+	return conn, err
+}
+
+// NewSystemConnection establishes a connection to the system bus and authenticates.
+// Callers should call Close() when done with the connection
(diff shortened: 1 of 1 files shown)


squeed:
Updated with suggestions from @lucab to prefer the system dbus versus the systemd dbus.

jonboulle:
Please rewrite the commit message to reflect the logic

dbus/dbus.go
```diff
@@ -83,9 +83,19 @@ type Conn struct {
...
 func New() (*Conn, error) {
+	conn, err := NewSystemConnection()
+	if err != nil && os.Geteuid() == 0 {
+		return NewSystemdConnection()
+	}
+	return conn, err
+}
+
+// NewSystemConnection establishes a connection to the system bus and authenticates.
+// Callers should call Close() when done with the connetion
```
jonboulle: connection

squeed:
@jonboulle updated, thanks! PTAL.

jonboulle:
Sorry to be That Guy, but I think the commit message could be improved a bit more :-). Something like:

```
dbus: make New() fall back to direct systemd connection

Previously, New() would attempt to connect to the 
system-wide DBus. This changes the behaviour so 
that in the event that such a connection fails, it falls 
back to attempting a direct connection to the private
systemd socket (i.e. via NewSystemdConnection()).

A new, more specific `NewSystemConnection()` constructor
is introduced which provides the previous behaviour.
```

After that LGTM

squeed:
Updated. Thanks!

PR #209, opened 2016-11-07
daemon: add watchdog checker

guilhem:
Add a function `SdWatchdogEnabled()` who check if service have to report its status
This is inspirated from the behavior of `sd_watchdog_enabled`

### DIFF
--- a/daemon/watchdog.go
+++ b/daemon/watchdog.go
@@ -1,0 +1,58 @@
+// Copyright 2016 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package daemon
+
+import (
+	"fmt"
+	"os"
+	"strconv"
+	"time"
+)
+
+// SdWatchdogEnabled return watchdog information for a service.
+// Process should send daemon.SdNotify("WATCHDOG=1") every time / 2.
+//
+// It returns one of the following:
+// (0, nil) - watchdog isn't enabled or we aren't the watched PID.
+// (0, err) - an error happened (e.g. error converting time).
+// (time, nil) - watchdog is enabled and we can send ping.
+//   time is delay before inactive service will be killed.
+func SdWatchdogEnabled() (time.Duration, error) {
+	wusec := os.Getenv("WATCHDOG_USEC")
+	if wusec == "" {
+		return 0, nil
+	}
+	s, err := strconv.Atoi(wusec)
+	if err != nil {
+		return 0, fmt.Errorf("error converting WATCHDOG_USEC: %s", err)
+	}
+	if s <= 0 {
+		return 0, fmt.Errorf("error WATCHDOG_USEC must a positive number")
+	}
+
+	wpid := os.Getenv("WATCHDOG_PID")
+	if wpid == "" {
+		return 0, nil
+	}
+	p, err := strconv.Atoi(wpid)
+	if err != nil {
+		return 0, fmt.Errorf("error converting WATCHDOG_PID: %s", err)
+	}
+	if os.Getpid() != p {
+		return 0, nil
+	}
+
+	return time.Duration(s) * time.Microsecond, nil
+}
--- a/daemon/watchdog_test.go
+++ b/daemon/watchdog_test.go
@@ -1,0 +1,122 @@
+// Copyright 2016 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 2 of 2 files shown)


guilhem:
I may need help to write tests on this feature

squeed:
Can you add documentation for the method? Thanks!

guilhem:
@squeed should be done :)

guilhem:
an exemple of usage: https://github.com/guilhem/traefik/commit/1caf8d9d7e45dcc076aa8e584b07745f0402d5e2

squeed:
Great, thanks.
I think you can test this pretty simply by just calling `os.Setenv` in your testing code. No need to get too complicated. Just store the variables somewhere when you start, and restore them when you finish.

daemon/watchdog.go
```diff
@@ -0,0 +1,58 @@
...
+
+// SdWatchdogEnabled return watchdog information for a service.
+// Process should send daemon.SdNotify("WATCHDOG=1") every time / 2.
+//
+// It returns one of the following:
+// (0, nil) - watchdog isn't enabled or we aren't the watched PID.
+// (0, err) - an error happened (e.g. error converting time).
+// (time, nil) - watchdog is enabled and we can send ping.
+//   time is delay before inactive service will be killed.
+func SdWatchdogEnabled() (time.Duration, error) {
```
lucab: `sd_watchdog_enabled()` takes an additional `unset_environment` bool which will unconditionally unset the two environment variables. I think it is worth to have it here too.

lucab:
@guilhem thanks for PR. It looks fine overall and I think that you can follow @squeed suggestion regarding testing it.  Additionally, I've left inline a small request which you may consider for the next iteration of this PR.

daemon/watchdog.go
lucab: I personally think it is fine if the two are not aligned. @jonboulle @squeed may have a different opinion though.

jonboulle: I don't really see a reason to introduce the inconsistency; perhaps we could land this and follow up adding that variable? (or is it somehow more useful for this call than for sdnotify?)

guilhem:
@squeed tests are there :)

daemon/watchdog.go
```diff
@@ -0,0 +1,58 @@
...
+	wusec := os.Getenv("WATCHDOG_USEC")
+	if wusec == "" {
+		return 0, nil
+	}
+	s, err := strconv.Atoi(wusec)
+	if err != nil {
+		return 0, fmt.Errorf("Error converting WATCHDOG_USEC: %s", err)
```
jonboulle: please lower-case error strings (e.g. "error converting...")

daemon/watchdog.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	if s <= 0 {
+		return 0, fmt.Errorf("Error WATCHDOG_USEC should be positive")
```
jonboulle: s/Error WATCHDOG_USEC should be positive/WATCHDOG_USEC must be a positive number/

daemon/watchdog.go
guilhem: Done

daemon/watchdog_test.go
```diff
@@ -0,0 +1,122 @@
...
+	"os"
+	"strconv"
+	"testing"
+)
+
+func TestSdWatchdogEnabled(t *testing.T) {
+	// (time, nil)
+	err := os.Setenv("WATCHDOG_USEC", "100")
+	if err != nil {
+		panic(err)
```
euank: `t.Fatal` preferred over panic to halt a test early.

daemon/watchdog_test.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	err = os.Setenv("WATCHDOG_PID", strconv.Itoa(os.Getpid()))
+	if err != nil {
+		panic(err)
+	}
+
+	delay, err := SdWatchdogEnabled()
+	if delay == 0 || err != nil {
+		t.Errorf("TEST: Watchdog enabled FAILED")
```
euank: If we're okay with this test being 1.7+ only, we could use [subtests](https://blog.golang.org/subtests) to cleanup these repetitive tests and still have nice error reporting.

I don't know how much this project needs compatibility with older go versions.

Failing at using subtests, I'd still rather have a slice of cases to loop through than this, though It's not a blocking concern.

daemon/watchdog_test.go
jonboulle: I actually disagree on that one - my usual classification is:
- `t.Error` for test failures related to user code that can continue
- `t.Fatal` for test failures related to user code that cannot continue
- `panic` for test failures unrelated to user code e.g. in setup etc (very explicit that it's an environment failure/solar flare and not a code bug)

guilhem:
And so?....

daemon/watchdog.go
squeed: Yeah, I think @jonboulle is right - might as well add the parameter. Sorry for the runaround.

squeed:
Tests look great, thanks!

daemon/watchdog_test.go
lucab: Agreeing with @jonboulle here, let's keep it as is.

daemon/watchdog_test.go
lucab: subtests are perhaps a bit of a stretch, I agree on having the test-cases in a better tabular form if possible.

lucab:
It looks like we agree this is mostly ok as it is. The only remaining concerns here are:
 * implementing `unset_env` for watchdog and notify
 * re-arranging the tests to be in a tabular form

Neither of those are blockers, so we can either do another iteration here or land this first and followup with a cleaning PR, depending on @guilhem availability. I'd suggest to at least re-arrange the tests in the scope of this PR, and be ready for a change in the function signature happening soon later. @guilhem sounds ok?

guilhem:
@lucab I'm not very familiar with go tests and don't have many ideas to improve those.
But with new github feature, any maintainer is free to add more commits in my `watchdog` branch to improve this PR ^^

jonboulle:
I need to do something mind-numbing after the US election so I'll take a shot.

jonboulle:
@lucab ptal

lucab:
LGTM, thanks everybody. I'll follow up with the `unset_environment` breaking change.

PR #210, opened 2016-11-09
Add more machine1 support

joejulian:
Added GetImage, GetMachine, GetProperties, and rudimentary support for KillMachine.

### DIFF
--- a/machine1/dbus.go
+++ b/machine1/dbus.go
@@ -18,15 +18,19 @@ limitations under the License.
 package machine1
 
 import (
+	"fmt"
 	"os"
 	"strconv"
 
 	"github.com/godbus/dbus"
+	"syscall"
 )
 
 const (
 	dbusInterface = "org.freedesktop.machine1.Manager"
 	dbusPath      = "/org/freedesktop/machine1"
+	SIGRTMIN      = syscall.Signal(32)
+	SIGINT        = syscall.SIGINT
 )
 
 // Conn is a connection to systemds dbus endpoint.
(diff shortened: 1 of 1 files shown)


machine1/dbus.go
```diff
@@ -75,6 +77,75 @@ func (c *Conn) initConnection() error {
...
+	machine = make(map[string]interface{}, len(props))
+	for k, v := range props {
+		machine[k] = v.Value()
+	}
+	return machine, nil
+}
+
+// KillMachine signals systemd to power off the container
+func (c *Conn) KillMachine(name string) error {
+	var killsignal int32 = 38 // SIGTERMIN+4 (systemd power off)
```
jonboulle: where do you get this number from? I can't seem to find it anywhere

machine1/dbus.go
```diff
@@ -75,6 +77,75 @@ func (c *Conn) initConnection() error {
 	return nil
 }
 
+// GetImage may be used to get the image object path for the image with the specified name
+func (c *Conn) GetImage(name string) (dbus.ObjectPath, error) {
+	var path dbus.ObjectPath
+	result := c.object.Call(dbusInterface+".GetImage", 0, name)
+	if result.Err != nil {
+		log.Print(result.Err)
```
jonboulle: Please remove this and other log lines

machine1/dbus.go
```diff
@@ -75,6 +77,75 @@ func (c *Conn) initConnection() error {
...
+		return path, result.Err
+	}
+	pathstr, ok := result.Body[0].(dbus.ObjectPath)
+	if ok {
```
jonboulle: Generally we prefer a) short-circuit returns, and b) explicitly returning `(nil, err)` in error cases - which also simplifies the code a bit:

```
func (c *Conn) GetImage(name string) (dbus.ObjectPath, error) {
	result := c.object.Call(dbusInterface+".GetImage", 0, name)
	if result.Err != nil {
		return nil, result.Err
	}
	path, ok := result.Body[0].(dbus.ObjectPath)
	if !ok {
		return nil, fmt.Errorf("Error converting '%v' to dbus.ObjectPath.", result.Body[0])
 	}
 	return path, nil
 }
```

jonboulle: (Same thing for GetMachine below)

jonboulle: (And for GetProperties :-))

joejulian: nil isn't an ObjectPath so it failed to compile

machine1/dbus.go
joejulian: I stole it from machinectl.c

I expect future enhancements to allow setting the kill signal, but for my immediate needs this is sufficient.

machine1/dbus.go
```diff
@@ -75,6 +76,72 @@ func (c *Conn) initConnection() error {
...
+		path = pathstr
+	} else {
+		return machine, fmt.Errorf("Error converting '%v' to dbus.ObjectPath.", result.Body[0])
+	}
+	obj := c.conn.Object("org.freedesktop.machine1", path)
+	err := obj.Call("org.freedesktop.DBus.Properties.GetAll", 0, "").Store(&props)
+	if err != nil {
+		return machine, err
```
purpleidea: these should return nil, err I think...

machine1/dbus.go
```diff
@@ -21,6 +21,7 @@ import (
 	"os"
 	"strconv"
 
+	"fmt"
```
purpleidea: Why is this down here instead of up a line?

machine1/dbus.go
```diff
@@ -75,6 +76,72 @@ func (c *Conn) initConnection() error {
...
+func (c *Conn) GetProperties(name string) (map[string]interface{}, error) {
+	var machine map[string]interface{}
+	var path dbus.ObjectPath
+	var props map[string]dbus.Variant
+	result := c.object.Call(dbusInterface+".GetMachine", 0, name)
+	if result.Err != nil {
+		return machine, result.Err
+	}
+	pathstr, ok := result.Body[0].(dbus.ObjectPath)
+	if ok {
```
purpleidea: as @jonboulle mentioned, the  inlining. Eg: https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

machine1/dbus.go
joejulian: No idea. Probably IntelliJ's auto-include. Fixed.

machine1/dbus.go
joejulian: which won't work in this case because of my choice of return types - which seemed logical since that's what the dbus method returns.

machine1/dbus.go
joejulian: Like I mentioned earlier, nil isn't a dbus.PathObject. I'm not hell-bent on keeping it that way if someone more familiar with the aim of the project thinks that's wrong. I was trying to be consistent with the dbus model.

purpleidea: @joejulian This is what I meant: https://play.golang.org/p/0X-DKJ5mht -- Did I miss something?

joejulian:
This branch has been taken over by @purpleidea. Closing PR.

PR #211, opened 2016-11-09
daemon: introduce support for cleaning env variables

lucab:
This fixes `SdWatchdogEnabled()` behavior when `$WATCHDOG_PID` is unset,
and introduces a new `unsetEnvironment` boolean flag to `SdWatchdogEnabled()`
and `SdNotify()` to clear releveant env variables, aligning to the behavior of native methods.

### DIFF
--- a/daemon/sdnotify.go
+++ b/daemon/sdnotify.go
@@ -22,11 +22,14 @@ import (
 )
 
 // SdNotify sends a message to the init daemon. It is common to ignore the error.
+// If `unsetEnvironment` is true, the environment variable `NOTIFY_SOCKET`
+// will be unconditionally unset.
+//
 // It returns one of the following:
 // (false, nil) - notification not supported (i.e. NOTIFY_SOCKET is unset)
 // (false, err) - notification supported, but failure happened (e.g. error connecting to NOTIFY_SOCKET or while sending data)
 // (true, nil) - notification supported, data has been sent
-func SdNotify(state string) (sent bool, err error) {
+func SdNotify(unsetEnvironment bool, state string) (sent bool, err error) {
 	socketAddr := &net.UnixAddr{
 		Name: os.Getenv("NOTIFY_SOCKET"),
 		Net:  "unixgram",
@@ -35,6 +38,13 @@ func SdNotify(state string) (sent bool, err error) {
 	// NOTIFY_SOCKET not set
 	if socketAddr.Name == "" {
 		return false, nil
+	}
+
+	if unsetEnvironment {
+		err = os.Unsetenv("NOTIFY_SOCKET")
+	}
+	if err != nil {
+		return false, err
 	}
 
 	conn, err := net.DialUnix(socketAddr.Net, nil, socketAddr)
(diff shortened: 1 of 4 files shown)


lucab:
@guilhem when checking the manpage for behavior of unset, I realized that https://github.com/coreos/go-systemd/pull/209 may have got the "missing $WATCHDOG_PID" case wrong. Can you please have a pass on this?

guilhem:
@lucab seems to be true :/
I've been trumped by https://github.com/Merovius/systemd/blob/master/watchdog.go#L26

daemon/sdnotify.go
```diff
@@ -22,20 +22,28 @@ import (
...
 // (false, nil) - notification not supported (i.e. NOTIFY_SOCKET is unset)
 // (false, err) - notification supported, but failure happened (e.g. error connecting to NOTIFY_SOCKET or while sending data)
 // (true, nil) - notification supported, data has been sent
-func SdNotify(state string) (sent bool, err error) {
+func SdNotify(unsetEnvironment bool, state string) (sent bool, err error) {
 	socketAddr := &net.UnixAddr{
 		Name: os.Getenv("NOTIFY_SOCKET"),
 		Net:  "unixgram",
 	}
+	err = os.Unsetenv("NOTIFY_SOCKET")
```
guilhem: Why  `unsetEnvironment` isn't check here ?

lucab:
I did two really silly mistakes in the first iteration of this, so both the code and the logic to test the unset feature were wrong :disappointed:. Rebased, @guilhem PTAL.

daemon/watchdog_test.go
```diff
@@ -30,29 +30,30 @@ func must(err error) {
...
+		{"1", "0", false, false, 0}, // WATCHDOG_PID doesn't match
+		{"", "", true, false, 0},    // Both not set
 
 		// Failure cases
-		{"-1", mypid, true, 0},                // Negative USEC
-		{"string", "1", true, 0},              // Non-integer USEC value
-		{"1", "string", true, 0},              // Non-integer PID value
-		{"stringa", "stringb", true, 0},       // E v e r y t h i n g
-		{"-10239", "-eleventythree", true, 0}, // i s   w r o n g
+		{"-1", mypid, true, true, 0},                // Negative USEC
```
jonboulle: is this mixing of true/false for unsetEnv arbitrary?

lucab: Yes, just making sure that it works. Unless we have any specific corner-case we want to cover, in which case better perhaps to align them and introduce another tests group below.

jonboulle: I guess it's fine - I'd just add a comment mentioning it

lucab: Ack, added a note next to the field.

jonboulle:
one comment, lgtm after that

daemon/sdnotify.go
```diff
@@ -22,20 +22,30 @@ import (
...
 // (false, nil) - notification not supported (i.e. NOTIFY_SOCKET is unset)
 // (false, err) - notification supported, but failure happened (e.g. error connecting to NOTIFY_SOCKET or while sending data)
 // (true, nil) - notification supported, data has been sent
-func SdNotify(state string) (sent bool, err error) {
+func SdNotify(unsetEnvironment bool, state string) (sent bool, err error) {
 	socketAddr := &net.UnixAddr{
 		Name: os.Getenv("NOTIFY_SOCKET"),
 		Net:  "unixgram",
 	}
+	if unsetEnvironment {
```
lucab: The order shouldn't matter in practice and I don't have any strong opionin about this, so I just swapped them around as suggested.

jonboulle:
LGTM, thanks!

PR #212, opened 2016-11-18
dbus: test: fix bugs and add more tests for unit properties

dongsupark:
1. dbus: close DBUS connection at the end of every test

As it's recommended to close every DBUS connection at the end of every test, let's add `conn.Close()` for each test. That will prevent potential bugs hitting an upper limit of connections on `/var/run/dbus/system_bus_socket`.

2. dbus: add more tests for unit properties

Add a common test `TestStartStopTransientUnit` for testing more properties to be used for transient units. The following properties are covered.

* PropType("oneshot")
* PropRemainAfterExit
* PropRequires, PropRequiresOverridable
* PropRequisite, PropRequisiteOverridable
* PropWants
* PropBindsTo
* PropConflicts

### DIFF
--- a/dbus/methods_test.go
+++ b/dbus/methods_test.go
@@ -16,7 +16,6 @@ package dbus
 
 import (
 	"fmt"
-	"math/rand"
 	"os"
 	"os/exec"
 	"path"
@@ -28,6 +27,11 @@ import (
 
 	"github.com/godbus/dbus"
 )
+
+type TrUnitProp struct {
+	name  string
+	props []Property
+}
 
 func setupConn(t *testing.T) *Conn {
 	conn, err := New()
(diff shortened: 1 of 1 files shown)


dongsupark:
I just excluded the 2nd patch that stopped the transient scope, as it caused travis to fail. It didn't fail on Fedora though. I'll try that patch later again with another PR.

dbus/methods_test.go
```diff
@@ -739,9 +754,527 @@ func TestStartStopTransientUnit(t *testing.T) {
 	}
 }
 
+// Ensure that oneshot transient unit starting and stopping works.
+func TestStartStopTransientUnitOneshot(t *testing.T) {
```
jonboulle: initial eyeballing suggests a lot of copy-paste between these test functions - have you tried to parameterise it into a test table?

dongsupark: @jonboulle Yeah, I also think it's too many duplications. Though I just couldn't come up with a clean abstraction between the test cases that differ slightly with each other. Let me think about how to simplify it.

dongsupark: @jonboulle I rewrote the 2nd patch, and force-pushed. With a test table, it will be more readable than before. PTAL.

lucab:
This slipped a bit through my backlog. @dongsupark @jonboulle what's the status here?

dongsupark:
@lucab I'm just waiting for review. If anything needs to be done, please let me know.

lucab:
Thanks a lot for enhacing the testsuite and sorry for the delay. Everything looks fine, thus merging.

dongsupark:
@lucab Thanks! ;-)

Issue #213, opened 2016-11-21
Additional Machined functionality


cpg1111:
Hi there,

I was taking a looking at go-systemd's machined functionality and I was wondering how come there's only `RegisterMachine()`?  

If adding additional functionality in there is something desired, I'd be happy to submit a pull request adding features such as GetMachine, GetImage, ListMachines, TerminateMachine, etc.  And if it'd be preferred that I add each one in a separate Pull Request, I can do that as well.

Thanks!

jonboulle:
There was actually a PR put up recently to add this but it was closed in favour of @purpleidea following up with a reworked version: https://github.com/coreos/go-systemd/pull/210

Feel free to use that one as a starting point, or just creating your own, but probably best to check with @purpleidea

purpleidea:
@jonboulle @cpg1111 I'm actually not personally working on this, I forked @joejulian's version because it was referenced in https://github.com/purpleidea/mgmt/ and changing his PR kept breaking our git master. I'm not sure if he's still working on the upstream patch or not, so I'd ask him, or go for it.

If you'd like to patch mgmt's nspawn resource with your improvements, that would be welcome too!

Cheers

joejulian:
Feel free to take what you want from https://github.com/joejulian/go-systemd/tree/machine1_support and create your own PR. I found some issues with race conditions within systemd and/or systemd-machined that interfere with the work I was doing on mgmt.

Additionally there seemed to be some sort of philosophical problem with the work I was doing and I'm novice enough with golang that I didn't have any idea how to make it acceptable.

cpg1111:
@purpleidea Ah ok.  I have already submitted Pull Request in this repo here https://github.com/coreos/go-systemd/pull/214, it's failing tests due to the need for root to create machines, which I discuss in the PR.  I hadn't come across https://github.com/purpleidea/mgmt/ before, I'll take a look and see if I can submit a Pull Request for that as well.

purpleidea:
On Mon, Nov 21, 2016 at 6:57 PM, Christian Grabowski
<notifications@github.com> wrote:
>
> @purpleidea Ah ok. I have already submitted Pull Request in this repo here #214, it's failing tests due to the need for root to create machines, which I discuss in the PR. I hadn't come across https://github.com/purpleidea/mgmt/ before, I'll take a look and see if I can submit a Pull Request for that as well.

Great! The WIP nspawn code is in resources/nspawn.go Feel free to ping
[...]

PR #214, opened 2016-11-21
Add Additional Machined Functions

cpg1111:
This is in regards to issue https://github.com/coreos/go-systemd/issues/213  and I'm aware there was an existing closed PR for this, but I figure this addresses some of those concerns that closed that PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ env:
 
 before_install:
  - docker pull ${DOCKER_BASE}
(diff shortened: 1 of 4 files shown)


cpg1111:
Tests are failing on a `docker exec`, does anyone happen to have some insight on this?

cpg1111:
So the tests fail due to not running the tests as root, which is required for creating a machine.  Are there any objections to running this as root?

jonboulle:
Could you just move machine1 alongside dbus in the test script? We already only run those if root and this existing non-root machine1 test isn't really doing anything.

cpg1111:
@jonboulle certainly. I just wasn't sure there were already tests running as root. I'll move those shortly.

cpg1111:
Tests are passing now.  I had to add systemd-machined into the containers as a separate package, other than that, all changes are local to the systemd-machined functions.

jonboulle:
@cpg1111 thanks so much! since there was a bit of back and forth with all the commits, I've rebased and made some tweaks and will try to land via #217

PR #221, opened 2017-02-01
test/journal: add some send tests

lucab:
This commit introduces tests exercising sending logic for
empty/small/large messages.

### DIFF
--- a/journal/journal_test.go
+++ b/journal/journal_test.go
@@ -15,6 +15,10 @@
 package journal
 
 import (
+	"fmt"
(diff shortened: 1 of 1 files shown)


lucab:
This is to test logic touched by https://github.com/coreos/go-systemd/pull/218. It seems to be currently buggy.

lucab:
@squeed I think we can merge this now, and uncomment the failing tests once Go 1.11 and #5 can be fixed in all supported toolchains. PTAL.

PR #224, opened 2017-02-19
sdjournal: fix goroutine leak

### LINKED ISSUE #222: sdjournal: goroutines stuck on channel send
Hi!

I have a problem with https://github.com/coreos/go-systemd/blob/e97b35f834b17eaa82afe3d44715c34736bfa12b/sdjournal/read.go#L218

I'm using a [slightly modified version](https://github.com/mheese/journalbeat/blob/652d30b90adde3e8406f308ee0a78fed7594c0a7/journal/follow.go#L86) of `(r *JournalReader) Follow()` returning a channel of `*JournalEntry`s but the waiting logic is preserved so I'm writing here.

I noticed that after some time the overall number of active goroutines grows slowly and differs from one host to another:
```
goroutine profile: total 29
17 @ 0x42ef3a 0x42f02e 0x406038 0x405dfd 0x4fa712 0x45fc51
#	0x4fa711	github.com/mheese/journalbeat/journal.Follow.func2.1+0x91	/github.com/mheese/journalbeat/journal/follow.go:86
```

Here is an example of a single goroutine dump:
```
goroutine 180549 [chan send, 16014 minutes]:
github.com/mheese/journalbeat/journal.Follow.func2.1(0xc4215d8230, 0xc4215d81c0, 0xc420177920)
	/github.com/mheese/journalbeat/journal/follow.go:86 +0x92
created by github.com/mheese/journalbeat/journal.Follow.func2
	/github.com/mheese/journalbeat/journal/follow.go:89 +0x29e
```

Here is a rough repro: https://play.golang.org/p/PNGLjD_S5m
When using a buffered channel of size 1 you are likely to get one event already sent to channel buffer, and one event pending to be sent to the channel before you get a `pollDone` signal. This causes the aforementioned issue with orphaned goroutines being stuck.

A couple of questions in this regard:
1. Is there any rationale behind handling a blocking "wait for new events to arrive" operation through spawning a goroutine with consequent `select`?
2. If there is any then why use buffered channels? And why use `default: events <- ...`? Both of them being reasons for a leak.

nicorevin:
Fixes #222

### DIFF
--- a/sdjournal/read.go
+++ b/sdjournal/read.go
@@ -185,6 +185,10 @@ func (r *JournalReader) Follow(until <-chan time.Time, writer io.Writer) (err er
 	// Process journal entries and events. Entries are flushed until the tail or
 	// timeout is reached, and then we wait for new events or the timeout.
 	var msg = make([]byte, 64*1<<(10))
+	var waitCh = make(chan int)
+	var waitStop = make(chan bool)
+	defer close(waitStop)
+
 process:
 	for {
 		c, err := r.Read(msg)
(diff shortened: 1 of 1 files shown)


lucab:
ok to test

sdjournal/read.go
```diff
@@ -196,43 +200,38 @@ process:
 		case <-until:
 			return ErrExpired
 		default:
-			if c > 0 {
-				if _, err = writer.Write(msg[:c]); err != nil {
-					break process
-				}
-				continue process
+		}
```
lucab: I'm not sure about the rationale behind this change.

nicorevin: just makes code less nested.

nicorevin:
Hi @lucab,

[this code](https://github.com/mheese/journalbeat/blob/12f42c21ee002537f8906ab2a4d4d4a59b175b91/journal/follow.go#L79-L100) has been in test for more than two weeks on more than 100 nodes. No leaks so far.
If you need more details and explanation or have any questions please let me know.

sdjournal/read.go
```diff
@@ -185,6 +185,10 @@ func (r *JournalReader) Follow(until <-chan time.Time, writer io.Writer) (err er
 	// Process journal entries and events. Entries are flushed until the tail or
 	// timeout is reached, and then we wait for new events or the timeout.
 	var msg = make([]byte, 64*1<<(10))
+	var waitCh = make(chan int)
+	var waitStop = make(chan bool)
```
lucab: If I'm not mistaken, these used to be 1-slot buffered channels. I think the change is on purpose to synchronize sender and receiver, right? Are there other hidden behavioral changes behind this?

lucab:
@nicorevin sorry, I haven't had time to properly review this so far. If I'm correctly grasping the changes, this PR looks fine. I've just left a minor question inline.

sdjournal/read.go
nicorevin: right. No other hidden changes. The main idea of this change is to synchronise sender and receiver to guarantee that only one goroutine is active at a time and use the inner for-loop for waiting for the new entries instead of using the whole `process` loop.

lucab: Ack, thanks for confirming.

PR #220, opened 2017-01-31
machine1: use random labels in tests

lucab:
This commit introduces a random suffix in machined tests, to make sure
that there are no conflicts between parallel testsuite runs.

PR #225, opened 2017-03-02
dbus: Add systemd user session

purpleidea:


purpleidea:
Note: I haven't fully tested this yet, but I think I'm on the right track. LMK

purpleidea:
Tested patch and it seems to work. The failure looks like it's unrelated AFAICT... Please verify.

purpleidea:
@jonboulle I suppose I'm suppose to ping someone for a review + merge.

Thanks!

lucab:
ok to test

jonboulle:
Code looks OK but tests are faulting for some reason

lucab:
I re-triggered master and it is now failing, however it was fine 29 days ago after merging https://github.com/coreos/go-systemd/pull/220 and nothing changed in the meanwhile.

lucab:
As a library, we are using godbus without vendoring/pinning to any specific version. Tests seem to be failing after https://github.com/godbus/dbus/pull/77 got merged.

jsouthworth:
I'll look into the godbus related failure today.

purpleidea:
As an aside, travis has built-in "cron" support to automatically rebuild master every day or so to hopefully catch these things earlier. It's easy to turn on in the settings: https://docs.travis-ci.com/user/cron-jobs/

jsouthworth:
I have the PR open that addresses this problem. I've checked tested this project against those changes:
ok  	github.com/coreos/go-systemd/dbus	3.524s

purpleidea:
Amended date on commit to re-trigger builds. AFAICT should be ready for merge now.

purpleidea:
@lucab Can you merge this? Thx.

lucab:
@purpleidea I was looking at the mgmt commit (which seems to be the main usecase behind this) and I'm not sure why `NewUserConnection` isn't enough in that context? What's the case where there is a systemd user session but no dbus available?

I'm asking because strictly speaking `/private` is still a private interface which may break at any point, so I'd avoid exposing it unless strictly needed.

purpleidea:
@lucab So I'm not completely familiar with all of the internals, so maybe I'm not doing this optimally! But I wanted to manage systemd services that are within the `--user` set, and/or what's seen with `systemd-run`. There seems to be a bit of overlap between those two which isn't entirely clear to me though.

lucab:
Then `NewUserConnection` should give you access to both transient and persistent units under the user session. The only difference compared to your `NewSystemdUserConnection` is that it relies on a dbus session bus, instead of going through the private socket.

Let me know if you are hitting some specific case where this doesn't work, so we can have a better look at it.

purpleidea:
On Mon, Mar 6, 2017 at 6:26 AM, Luca Bruno <notifications@github.com> wrote:
>
> Then NewUserConnection should give you access to both transient and persistent units under the user session. The only difference compared to your NewSystemdUserConnection is that it relies on a dbus session bus, instead of going through the private socket.
>
> Let me know if you are hitting some specific case where this doesn't work, so we can have a better look at it.

Is the only difference between --user transient and non transient the
location of the .service files? And is that the difference between
$XDG_RUNTIME_DIR/systemd/user/ and ~/.config/systemd/user/ ?

Or is something else going on? And is there a way to create
non-transient user services with systemd-run?

Thanks for your help!

lucab:
> Is the only difference between --user transient and non transient the location of the .service files? And is that the difference between $XDG_RUNTIME_DIR/systemd/user/ and ~/.config/systemd/user/ ? Or is something else going on?

That should be mostly it (plus some obscure interactions/constraints with generators, which I'm not even completely sure about).

> And is there a way to create non-transient user services with systemd-run?

Not that I know.

purpleidea:
@lucab Thanks for the discussion! Can we still merge this? Thanks!

lucab:
> Then NewUserConnection should give you access to both transient and persistent units under the user session.

> Let me know if you are hitting some specific case where this doesn't work, so we can have a better look at it.

To expand my previous comment, I think this PR is not needed at all and I would refrain from exposing this connection to the private user-session socket unless it covers some specific case which I missed.

purpleidea:
@lucab Hey just saw this now. I wrote an added patch to the set before I read this.

Can you tell me why others have been using the private interface (eg: why it was there before?) I have had good performance with it in mgmt, but I have no yet had a chance to try the `normal` method.

If you are still interested in this patch, please have a quick look at the new commit.

Thanks!

lucab:
> Can you tell me why others have been using the private interface (eg: why it was there before?)

For the system bus, because you may need to talk to systemd manager before the dbus-daemon is running (or in constrained environment where there is none at all). This doesn't seem to me to be a concern for the user session, which is why I think it should not be abused by the library.

purpleidea:
@lucab (Sorry for the delay in my reply.)

So in the common case for normal user accounts, I would completely agree, however I've been playing with this as a view to taking existing system services (such as Glusterd which itself spawns multiple children processes) and instead has it break out and ask systemd to spawn them for it. This gives us the obvious benefit of getting cgroups per sub-service, and everything else that systemd offers.

Doing this early would be advantageous, but I can't guarantee that it's necessary at this point. So I think it's reasonable to have this merged, but I can't *prove* that I need it yet.

I added a new patch which makes using this more sane, and only falls back to the private method if there is an issue.

jonboulle:
@purpleidea I'm struggling to grok your last message; would you mind walking us through an example?

purpleidea:
@jonboulle Sorry about that!

Okay currently Glusterd does this:

`systemd -> glusterd -> glusterFSd{1..N}`

all the individual brick processes are under the same service of `glusterd`, and it builds in it's own process management which isn't ideal.

One new possible plan is this:

`systemd -> gd3 (new glusterd) -> asks systemd to add new user services`
`-> receives ask to systemd -> glusterFSd{1..N}`

Initially when writing the patch I didn't realize all the semantics of the `/private` but @lucab was very kind in explaining it all. So now my best approach is to use the `UserNew()` function I added in the second part of the patch which falls back to `dbus-daemon` normally in cases where it might have crashed or is buggy. It's best to try and get `some` communication, rather than none when storage is involved, so I consider this safer.

HTH, please LMK if there are other questions, and thanks for the review!

jonboulle:
@purpleidea gotcha, thanks for the explanation! I think the use case is clear now. But just so I fully understand - this is a hypothesised idea but is not being implemented anywhere (yet)?

I'm hesitant to move forward with this because in the past when we've explored using this socket file we've been steered firmly away by the systemd maintainers:
https://github.com/systemd/systemd/issues/3610#issuecomment-229679807

The main argument I'd see for adding this is i) consistency (since we already unfortunately leak that private API; probably my fault for allowing https://github.com/coreos/go-systemd/issues/207 through); or ii) if this is demonstrably needed by extant projects. Thoughts?

purpleidea:
@jonboulle 

> this is a hypothesised idea but is not being implemented anywhere

Actually no, I have running (unfinished) code that uses this, but I haven't yet push it out publicly yet. (My bad, coming soon, was cleaning up a few things and also waiting for a final ACK/NACK on this first.)

RE: consistency, I completely agree, which is why I added the second patch so that I use `UserNew` instead, and that way it's stable if we take away the private part. I left the `NewSystemdUserConnection` public because I wanted to be able to experiment with it more. I could add a comment to the function doc that says this isn't part of the go-systemd stable API. Would that work?

lucab:
> waiting for a final ACK/NACK on this first

I think the general consensus is an overall-nack. In particular:

> in cases where it might have crashed or is buggy

This is not a proper usecase, as systemd / dbus-daemon / systemd-user-session will react to this in some ways (which I don't know by heart, but I imagine can entail checkpointing and chain-restarting) and having a library trying to bypass it via private connection just adds more failure points.

So I'd still stick to my original comment in https://github.com/coreos/go-systemd/pull/225#issuecomment-285287838 and I'd be happy to re-evaluate this conclusion if `NewUserConnection` proves not to be enough in some real usecases.

purpleidea:
@lucab Okay, thanks for the reviews, I'll close this then.

PR #228, opened 2017-03-16
test: Specify periodic go-systemd builder job as dsl

mxinden:
Instead of configuring the 'Periodic go-systemd builder' Jenkins job in
the Jenkins-UI this commit introduces the job specification via Jenkins
job dsl language.


mxinden:
@lucab Could you review this PR. 

The test execution is failing because the test script is not yet present on master.

Issue #145, opened 2016-03-17
timeout value passed to sd_journal_wait should be a duration, not absolute timestamp


PiersEBMS:
sd_journal_wait appears from the man page to take a (relative) timeout value but the code passes in an absolute wall-clock time, which will presumably be massively larger than what we meant

In sdjournal.Wait():
    to = uint64(time.Now().Add(timeout).Unix() / 1000)
should be
    to = uint64(timeout / 1000)

It certainly works better with this change at my end.

PR #231, opened 2017-03-23
sdjournal: fix Wait() timeout

phemmer:
This fixes a bug in the Wait() method on sdjournal.Journal{}.
The C API expects a duration in microseconds. However the method was passing a nonsense value (timeout's absolute time as thousands of seconds since epoch). This resulted in the value being capped to a maximum of 1.5s.

Looks like the bug was introduced here: https://github.com/coreos/go-systemd/commit/31020344488513bba6094ac4bd728c8d7f768194#diff-985ded8023b21268d0d5cf7f29781d2eR203

Fixes https://github.com/coreos/go-systemd/issues/145


lucab:
ok to test

PR #234, opened 2017-06-07
sdjournal: add NewJournalFromFiles

msackman:
Needed sd_journal_open_files. So added it.

### DIFF
--- a/sdjournal/journal.go
+++ b/sdjournal/journal.go
@@ -47,6 +47,15 @@ package sdjournal
 //   return sd_journal_open_directory(ret, path, flags);
 // }
 //
+// int
+// my_sd_journal_open_files(void *f, sd_journal **ret, const char **paths, int flags)
+// {
+//   int (*sd_journal_open_files)(sd_journal **, const char **, int);
+//
+//   sd_journal_open_files = f;
+//   return sd_journal_open_files(ret, paths, flags);
+// }
+//
 // void
 // my_sd_journal_close(void *f, sd_journal *j)
 // {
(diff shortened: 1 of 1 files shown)


msackman:
Well that's not fair! 503 when travis is trying to download stuff from debian stretch. Not even my code!

lucab:
ok to test

sdjournal/journal.go
```diff
@@ -426,6 +435,33 @@ func NewJournalFromDir(path string) (j *Journal, err error) {
 	return j, nil
 }
 
+// NewJournalFromFiles returns a new Journal instance pointing to a journals residing
+// in a given files. The supplied path may be relative or absolute; if
+// relative, it will be converted to an absolute path before being opened.
```
lucab: Where is this abs() applied? I didn't find it neither here nor mentioned in the `sd_journal_open` manpage.

lucab:
Thanks for this PR, it looks fine overall. I just have a minor doubt on the documented behavior. I re-triggered travis and it is green now. I also slightly amended the PR title, it would be nice to re-word the commit title as well.

msackman:
Thanks for the review. Yes, I was puzzled about the comment in the other function too but I'm afraid I just copied that over. Looking at `man 2 open`, it looks like libc actually copes with relative paths by prefixing $CWD. So I suppose the doc/comment is just referring to that behaviour.

lucab:
Then it looks like we are documenting some behavior which is actually an internal implementation several layers down. I suggest to either drop that part or to ensure that behavior by doing a conditional cwd-prefixing ourself.

msackman:
Perfectly agree. I'll remove both those comments in the docs then - will get to that early tomorrow. Cheers.

msackman:
Heh, same travis failure as before.

msackman:
Many thanks.

PR #239, opened 2017-09-26
sdjournal: support testing cursor against current position

adrianlzt:
This fixes TestCursor so it returns an error when the ``cursor``
parameter is not the same as the current position.

From SD_JOURNAL_GET_CURSOR(3):
```
sd_journal_test_cursor() returns positive if the current entry
matches the specified cursor, 0 if it does not match the specified
cursor or a negative errno-style error code on failure.
```

### DIFF
--- a/sdjournal/journal.go
+++ b/sdjournal/journal.go
@@ -303,6 +303,7 @@ package sdjournal
 import "C"
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


sdjournal/journal.go
```diff
@@ -879,6 +879,8 @@ func (j *Journal) TestCursor(cursor string) error {
 
 	if r < 0 {
 		return fmt.Errorf("failed to test to cursor %q: %d", cursor, syscall.Errno(-r))
+	} else if r == 0 {
+		return fmt.Errorf("Cursor parameter is not the same as current position")
```
lucab: Can you please change this into a top-level error constant and compare against it in the test?

adrianlzt:
``fmt.Errorf`` could not be defined as a constant.

lucab:
Yes sorry for the miswording it, golang doesn't support error constants. A top-level exported variable with a static value is the idiomatic way of getting that.

sdjournal/journal.go
```diff
@@ -370,6 +370,12 @@ const (
 	IndefiniteWait time.Duration = 1<<63 - 1
 )
 
+var (
+	// Error show when using TestCursor function and cursor parameter is not the same
+	// as the current cursor position
+	ErrNoTestCursor = fmt.Errorf("Cursor parameter is not the same as current position")
```
lucab: There is no formatting going on here, so `errors.New()` is enough.

lucab:
Thanks, looks good. If you could just squash all commits together, this is otherwise ready to go.

PR #238, opened 2017-08-07
Add in basic login session handling

paultag:
This is far from complete, but I'm starting to sketch out what Session
management could look like from login1.

PR #242, opened 2018-01-02
login1: add support for session management

siavashs:
This PR adds support for more logind methods, specially for session management.
Closes #238

### DIFF
--- a/login1/dbus.go
+++ b/login1/dbus.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Integration with the systemd logind API.  See http://www.freedesktop.org/wiki/Software/systemd/logind/
+// Package login1 provides integration with the systemd logind API.  See http://www.freedesktop.org/wiki/Software/systemd/logind/
 package login1
 
 import (
@@ -34,7 +34,7 @@ type Conn struct {
 	object dbus.BusObject
 }
 
-// New() establishes a connection to the system bus and authenticates.
+// New establishes a connection to the system bus and authenticates.
 func New() (*Conn, error) {
 	c := new(Conn)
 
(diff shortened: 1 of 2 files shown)


squeed:
Looks good. Any idea on how to add tests that aren't fragile?

siavashs:
I think we can run `loginctl` and compare the result for some of the methods.

login1/dbus.go
```diff
@@ -74,6 +74,115 @@ func (c *Conn) initConnection() error {
...
+func (s Session) toInterface() []interface{} {
+	return []interface{}{s.ID, s.UID, s.User, s.Seat, s.Path}
+}
+
+func sessionFromInterfaces(session []interface{}) (*Session, error) {
+	id := session[0].(string)
+	uid := session[1].(uint32)
+	user := session[2].(string)
+	seat := session[3].(string)
+	path := session[4].(dbus.ObjectPath)
```
lucab: Please perform checked type assertions here instead. That is, something like:
```
path, ok := session[4].(dbus.ObjectPath)
if !ok {
    return nil, fmt.Errorf("failed to typecast session field 4 to ObjectPath")
}
```

login1/dbus.go
```diff
@@ -74,6 +74,115 @@ func (c *Conn) initConnection() error {
...
+
+	ret := Session{ID: id, UID: uid, User: user, Seat: seat, Path: path}
+	return &ret, nil
+
+}
+
+func userFromInterfaces(user []interface{}) (*User, error) {
+	uid := user[0].(uint32)
+	name := user[1].(string)
+	path := user[2].(dbus.ObjectPath)
```
lucab: Same as above. Please add a length check and checked type assertions.

login1/dbus.go
```diff
@@ -74,6 +74,115 @@ func (c *Conn) initConnection() error {
...
+}
+
+// GetSession may be used to get the session object path for the session with the specified ID.
+func (c *Conn) GetSession(id string) (dbus.ObjectPath, error) {
+	var out interface{}
+	if err := c.object.Call(dbusInterface+".GetSession", 0, id).Store(&out); err != nil {
+		return "", err
+	}
+
+	return out.(dbus.ObjectPath), nil
```
lucab: Please add a checked type assertion for `out`.

login1/dbus.go
```diff
@@ -74,6 +74,115 @@ func (c *Conn) initConnection() error {
...
+
+// ListUsers returns an array with all currently logged in users.
+func (c *Conn) ListUsers() ([]User, error) {
+	out := [][]interface{}{}
+	if err := c.object.Call(dbusInterface+".ListUsers", 0).Store(&out); err != nil {
+		return nil, err
+	}
+
+	ret := []User{}
+
```
lucab: Spurious newline.

login1/dbus.go
```diff
@@ -74,6 +74,115 @@ func (c *Conn) initConnection() error {
...
+// ListSessions returns an array with all current sessions.
+func (c *Conn) ListSessions() ([]Session, error) {
+	out := [][]interface{}{}
+	if err := c.object.Call(dbusInterface+".ListSessions", 0).Store(&out); err != nil {
+		return nil, err
+	}
+
+	ret := []Session{}
+
```
lucab: Spurious newline.

lucab:
It would be nice to have at least some basic smoke tests calling those new methods and ensuring they return something meaningful without panicking. I think the docker setup on travis already has a working logind to query.

siavashs:
I'll look into tests.

login1/dbus.go
```diff
@@ -74,6 +74,123 @@ func (c *Conn) initConnection() error {
...
+func (s Session) toInterface() []interface{} {
+	return []interface{}{s.ID, s.UID, s.User, s.Seat, s.Path}
+}
+
+func sessionFromInterfaces(session []interface{}) (*Session, error) {
+	id := session[0].(string)
+	uid := session[1].(uint32)
+	user := session[2].(string)
+	seat := session[3].(string)
+	path, ok := session[4].(dbus.ObjectPath)
```
lucab: Sorry, I wasn't clear enough in my previous review. Please make sure that all type assertions (i.e. also fields 0 to 3) are checked, like you did for the last field.

login1/dbus.go
```diff
@@ -74,6 +74,123 @@ func (c *Conn) initConnection() error {
...
+	}
+
+	ret := Session{ID: id, UID: uid, User: user, Seat: seat, Path: path}
+	return &ret, nil
+}
+
+func userFromInterfaces(user []interface{}) (*User, error) {
+	uid := user[0].(uint32)
+	name := user[1].(string)
+	path, ok := user[2].(dbus.ObjectPath)
```
lucab: Same as above, and apologies that my previous review was not precise.

lucab:
Bump on this. It looks like some comments in my previous review were not precise enough, apologies for that.

siavashs:
No problem, I'll fix it.
I'll try to work on the tests this weekend.

lucab:
Just dropping a quick note, I'm planning to tag v16 in the next few days (once #248 and #249 are in). Unless you need this quickly in a release, we can pick this post-release once done, without hurrying.

siavashs:
As I checked there are no active user sessions inside a docker containers so logind responds with empty lists of users and sessions.
I'm looking into a simple way to create sessions inside a container, login1 provides `CreateSession` but it does not seem to be usable directly.
Let me know if you found any solution.

siavashs:
So I implemented the tests by creating and activating a test user and running the tests with the same user.
The results from `login1` are compared to values from `os/user.Current()`.

siavashs:
It seems running **sdjournal** tests as the new **test** user fails, while it works fine for me locally as an unprivileged user.
Any idea what might cause that error on the CI?

siavashs:
These latest changes on my branch [work on Ubuntu but fail on Debian](https://travis-ci.org/EclecticIQ/go-systemd/builds/337184330)!
But it fails for both in the PR tests.

login1/dbus.go
```diff
@@ -74,6 +74,141 @@ func (c *Conn) initConnection() error {
...
+	Name string
+	Path dbus.ObjectPath
+}
+
+func (s Session) toInterface() []interface{} {
+	return []interface{}{s.ID, s.UID, s.User, s.Seat, s.Path}
+}
+
+func sessionFromInterfaces(session []interface{}) (*Session, error) {
+	id, ok := session[0].(string)
```
lucab: Please check for `len(session) >= 5` before starting to index-access the array.

login1/dbus.go
```diff
@@ -74,6 +74,141 @@ func (c *Conn) initConnection() error {
...
+	if !ok {
+		return nil, fmt.Errorf("failed to typecast session field 4 to ObjectPath")
+	}
+
+	ret := Session{ID: id, UID: uid, User: user, Seat: seat, Path: path}
+	return &ret, nil
+}
+
+func userFromInterfaces(user []interface{}) (*User, error) {
+	uid, ok := user[0].(uint32)
```
lucab: Please check for `len(user) >= 3` before starting to index-access the array.

lucab:
@siavashs While I understand it would be nice to exact-match the environment in the test, I think that that is too fragile and also requires manipulating the environment first. I would suggest instead to be a bit less strict: just ensure that calls to `ListUsers` and `ListSessions` don't panic or return errors, and if there are actual users/sessions let's loop through them and ensure fields are populated.

login1/dbus.go
siavashs: Should we return an error if the number of fields don't match?

login1/dbus.go
siavashs: Should we return an error if the number of fields don't match?

login1/dbus.go
lucab: Yes. Rationale is that we prefer to return an error earlier to the caller if the input array is too short, instead of panicing in library code.

login1/dbus.go
lucab: Same as above, yes.

siavashs:
Maybe an unrelated question, but why do I see a higher coverage % when running tests locally?
```
...
ok  	github.com/coreos/go-systemd/login1	0.009s	coverage: 53.7% of statements
...
```
Is this related to go version?

lucab:
@siavashs my guess is that given that there are no sessions/users in the CI run, some codepaths within conditional paths are not reached. For example, both `*FromInterface` are only called inside a range iterator thus are not triggered on empty arrays.

This PR seems complete to me now, is there anything else you planned to add here? If not, can you please squash commits together and adopt the same title as the PR one?

siavashs:
@lucab it is ready 😄

lucab:
LGTM 💚 . Thanks for your contribution (and patience)!

PR #244, opened 2018-01-22
dbus: fix subStateSubscriber crash

dtnaylor:
When dispatching updates to sub-state subscribers, `sendSubStateUpdate`
previously called `GetUnitProperties` with a full, escaped unit path.
`getUnitProperties` expected only an unescaped unit name.

This commit changes `getUnitProperties` to take the full, escaped path,
updates `GetUnitProperties` to escape and expand the unit name to a
path, and adds `GetUnitPathProperties` for callers who already have the
full, escaped unit path.

To quickly see the problem, run this go program:
```
package main

import (
        "fmt"
        "github.com/coreos/go-systemd/dbus"
)

func main() {
        conn, _ := dbus.New()
        defer conn.Close()

        subUpdateCh := make(chan *dbus.SubStateUpdate)
        subErrCh := make(chan error)
        conn.SetSubStateSubscriber(subUpdateCh, subErrCh)

        for {
                select {
                case subStateUpdate := <-subUpdateCh:
                        fmt.Printf("%+v", subStateUpdate)
                case subStateErr := <-subErrCh:
                        fmt.Printf("error: %v", subStateErr)
                }
        }
}
```

With this program running, do something that updates a systemd unit. For example, if I run `sudo systemctl restart cron.service`, then I see the following output.

*With PR:*
```
&{UnitName:cron.service SubState:running}
```

*Without PR:*
```
panic: interface conversion: interface {} is nil, not string

goroutine 19 [running]:
github.com/coreos/go-systemd/dbus.(*Conn).sendSubStateUpdate(0xc4200c2620, 0xc4200ec060, 0x2d)
        /home/vagrant/go/src/github.com/coreos/go-systemd/dbus/subscription.go:195 +0x40c
github.com/coreos/go-systemd/dbus.(*Conn).dispatch.func1(0xc420064120, 0xc4200c2620)
        /home/vagrant/go/src/github.com/coreos/go-systemd/dbus/subscription.go:94 +0x10c
created by github.com/coreos/go-systemd/dbus.(*Conn).dispatch
        /home/vagrant/go/src/github.com/coreos/go-systemd/dbus/subscription.go:62 +0x83
exit status 2
```

Printing `err` from https://github.com/coreos/go-systemd/blob/master/dbus/subscription.go#L186 shows:
```
Unit name /org/freedesktop/systemd1/unit/cron_2eservice is not valid.
```

And printing `path` from https://github.com/coreos/go-systemd/blob/master/dbus/methods.go#L156 shows that we're winding up with the `/org/freedesktop/systemd1/unit/` part of the path repeated (the second time escaped):
```
/org/freedesktop/systemd1/unit/_2forg_2ffreedesktop_2fsystemd1_2funit_2fcron_5f2eservice
```

If I missed something and this isn't an issue, or if you'd like to fix it differently, I'm happy to close the PR and resubmit as an issue instead.

### DIFF
--- a/dbus/methods.go
+++ b/dbus/methods.go
@@ -16,6 +16,7 @@ package dbus
 
 import (
 	"errors"
+	"fmt"
 	"path"
 	"strconv"
 
@@ -148,14 +149,13 @@ func (c *Conn) ResetFailedUnit(name string) error {
 	return c.sysobj.Call("org.freedesktop.systemd1.Manager.ResetFailedUnit", 0, name).Store()
 }
 
-// getProperties takes the unit name and returns all of its dbus object properties, for the given dbus interface
-func (c *Conn) getProperties(unit string, dbusInterface string) (map[string]interface{}, error) {
+// getProperties takes the unit path and returns all of its dbus object properties, for the given dbus interface
+func (c *Conn) getProperties(path dbus.ObjectPath, dbusInterface string) (map[string]interface{}, error) {
 	var err error
 	var props map[string]dbus.Variant
 
-	path := unitPath(unit)
 	if !path.IsValid() {
-		return nil, errors.New("invalid unit name: " + unit)
+		return nil, fmt.Errorf("invalid unit name: %v", path)
 	}
 
 	obj := c.sysconn.Object("org.freedesktop.systemd1", path)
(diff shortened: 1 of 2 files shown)


dbus/methods.go
```diff
@@ -16,6 +16,8 @@ package dbus
 
 import (
 	"errors"
+	"fmt"
+	"log"
```
lucab: Spurious "log" import.

dbus/subscription.go
```diff
@@ -16,6 +16,7 @@ package dbus
 
 import (
 	"errors"
+	"log"
```
lucab: Spurious "log" import.

dtnaylor:
Removed the unused "log" imports.

lucab:
Thanks for this PR! I'd like to spend a bit more time digging into this, but I may have a bit of backlog at the moment. Where is the failing type-assertion located? Would you mind attaching the full stacktrace you are getting?

dtnaylor:
Great, thanks for looking into this! I updated the original post with more info. Let me know if you need anything else.

lucab:
@dtnaylor thanks, LGTM, your investigation was on spot. I'm merging it now for the bugfix, but if you have time I'll be happy to take a followup PR with a testcase for this (similar to your main, a `daemon-reload` should also trigger it). I also a few more cleanups that I'd like to add later, so if you don't mind I'll CC you on those.

dtnaylor:
Sure, will do. And happy to look at any cleanup when you get to it.

PR #245, opened 2018-01-24
login1: add close function

sdemos:
the close function closes the internal dbus connection, which is otherwise inaccessible. This allows for cleaner shutdown for clients of this library.

### DIFF
--- a/login1/dbus.go
+++ b/login1/dbus.go
@@ -45,6 +45,17 @@ func New() (*Conn, error) {
 	return c, nil
 }
 
+// Close closes the dbus connection
+func (c *Conn) Close() {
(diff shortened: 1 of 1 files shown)


login1/dbus.go
```diff
@@ -45,6 +45,11 @@ func New() (*Conn, error) {
 	return c, nil
 }
 
+// Close closes the dbus connection
+func (c *Conn) Close() {
+	c.conn.Close()
```
lucab: What about checking for `nil` on both `c` and `c.conn` so we are sure a deferred `c.Close()` never panics?

sdemos: fixed

PR #246, opened 2018-01-29
test: add test for dbus substate subscriptions

dtnaylor:
Add a test for the fix in
https://github.com/coreos/go-systemd/commit/e1db11dbbff25725abb85feff85529b76787d038


dtnaylor:
@lucab

lucab:
Thanks for the quick followup, merging.

PR #247, opened 2018-01-30
dbus: add SystemState method for getting system state

arnested:
Equivalent to `systemctl is-system-running`.


lucab:
Thanks for the PR! Code looks good, would you mind adding a test to cover this? Just something basic to check that we don't error or panic, and that the resulting property is the string representation for one of the sane states.

arnested:
Cool. I'll add some testing later today or tomorrow!

Thank you.

arnested:
@lucab, please have a look now. I have added a test case.

lucab:
Thanks, LGTM.

PR #248, opened 2018-01-31
dbus: add SetPropertiesSubscriber method

dtnaylor:
This PR adds a `SetPropertiesSubscriber` method which is similar to
`SetSubStateSubscriber` but with two important differences:

* Each update includes all changed properties, not just SubState
* Each set of property values from systemd is reported; no update can be
  missed. With `SubStateSubscriber`, transient states can be missed
  because `sendSubStateUpdate` calls `GetUnitPathProperties` after receiving
  the original signal from systemd; by this point, the unit's state may
  have changed again. (This behavior is clearly documented, but for some
  use cases it might not be acceptable to miss a state transition.)

### DIFF
--- a/dbus/dbus.go
+++ b/dbus/dbus.go
@@ -16,6 +16,7 @@
 package dbus
 
 import (
+	"encoding/hex"
 	"fmt"
 	"os"
 	"strconv"
@@ -53,6 +54,27 @@ func PathBusEscape(path string) string {
 		if needsEscape(i, c) {
 			e := fmt.Sprintf("_%x", c)
 			n = append(n, []byte(e)...)
+		} else {
+			n = append(n, c)
+		}
+	}
+	return string(n)
+}
+
+// pathBusUnescape is the inverse of PathBusEscape.
+func pathBusUnescape(path string) string {
+	if path == "_" {
+		return ""
+	}
+	n := []byte{}
+	for i := 0; i < len(path); i++ {
+		c := path[i]
+		if c == '_' && i+2 < len(path) {
+			res, err := hex.DecodeString(path[i+1 : i+3])
+			if err == nil {
+				n = append(n, res...)
+			}
+			i += 2
 		} else {
 			n = append(n, c)
 		}
--- a/dbus/dbus.go
+++ b/dbus/dbus.go
@@ -96,12 +96,17 @@ type Conn struct {
 		jobs map[dbus.ObjectPath]chan<- string
 		sync.Mutex
 	}
-	subscriber struct {
+	subStateSubscriber struct {
 		updateCh chan<- *SubStateUpdate
 		errCh    chan<- error
 		sync.Mutex
 		ignore      map[dbus.ObjectPath]int64
 		cleanIgnore int64
+	}
+	propertiesSubscriber struct {
+		updateCh chan<- *PropertiesUpdate
+		errCh    chan<- error
+		sync.Mutex
 	}
 }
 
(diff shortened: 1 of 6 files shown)


dbus/dbus.go
```diff
@@ -60,6 +61,27 @@ func PathBusEscape(path string) string {
...
+// PathBusUnescape is the inverse of PathBusEscape.
+func PathBusUnescape(path string) string {
+	if path == "_" {
+		return ""
+	}
+	n := []byte{}
+	for i := 0; i < len(path); i++ {
+		c := path[i]
+		if c == '_' {
+			res, err := hex.DecodeString(path[i+1 : i+3])
```
lucab: Before looking ahead, please check that `i+3` does not overflow `len(path)`.

dbus/dbus.go
```diff
@@ -60,6 +61,27 @@ func PathBusEscape(path string) string {
 	return string(n)
 }
 
+// PathBusUnescape is the inverse of PathBusEscape.
+func PathBusUnescape(path string) string {
```
lucab: Do you need this to be public in order to use it in your consuming application? If not, I'd prefer to keep it private for the moment, we can always export it later if required.

dbus/subscription.go
```diff
@@ -84,6 +85,7 @@ func (c *Conn) dispatch() {
 			case "org.freedesktop.DBus.Properties.PropertiesChanged":
 				if signal.Body[0].(string) == "org.freedesktop.systemd1.Unit" {
 					unitPath = signal.Path
+					c.sendPropertiesUpdate(unitPath, signal.Body[1].(map[string]dbus.Variant))
```
lucab: Please add a `len(signal.Body) >= 2` check and a type-check for `signal.Body[1].(map[string]dbus.Variant)`.

dbus/subscription.go
```diff
@@ -231,20 +237,61 @@ func (c *Conn) updateIgnore(path dbus.ObjectPath, info map[string]interface{}) {
...
+		return
+	}
+
+	update := &PropertiesUpdate{unitName(unitPath), changedProps}
+
+	select {
+	case c.propertiesSubscriber.updateCh <- update:
+	default:
+		select {
+		case c.propertiesSubscriber.errCh <- errors.New("update channel full!"):
```
lucab: "update channel is full", no need for exclamation marks 😃 .

dbus/subscription.go
```diff
@@ -231,20 +237,61 @@ func (c *Conn) updateIgnore(path dbus.ObjectPath, info map[string]interface{}) {
...
+		default:
+		}
```
lucab: If the linter doesn't complain about a spurious return (I didn't try), perhaps consider adding an explicit `return` after the inner select block. That way, code eventually added at the bottom is already fine with error handling.

lucab:
Thanks for your PR, I left a few comments. This will probably conflict with https://github.com/coreos/go-systemd/pull/249, thus depending on timings you may need to rebase at some later point.

Can you add at least a simple test to exercise `SetPropertiesSubscriber` codepath?

dbus/dbus.go
dtnaylor: Nope, will make private.

dtnaylor:
Ok, those fixups should address your comments. I'm happy to squash and rebase once #249 is in if this looks good.

lucab:
Thanks, I'm fine with this. Just squash and rebase your commits, I'll take care of merging this before #249 and then solve conflicts on my side.

dtnaylor:
Great—squashed and rebased.

PR #249, opened 2018-01-31
dbus: make SetSubStateSubscriber safer

lucab:
This adds several nil and type checks in the dbus subscription logic,
to ensure it won't panic in case of unexpected events.

### DIFF
--- a/dbus/subscription.go
+++ b/dbus/subscription.go
@@ -16,6 +16,7 @@ package dbus
 
 import (
 	"errors"
+	"log"
 	"time"
 
 	"github.com/godbus/dbus"
@@ -176,6 +177,16 @@ type SubStateUpdate struct {
 // is full, it attempts to write an error to errCh; if errCh is full, the error
 // passes silently.
 func (c *Conn) SetSubStateSubscriber(updateCh chan<- *SubStateUpdate, errCh chan<- error) {
+	if c == nil {
+		msg := "nil receiver"
+		select {
+		case errCh <- errors.New(msg):
+		default:
+			log.Printf("full error channel while reporting: %s\n", msg)
+		}
+		return
+	}
+
 	c.subStateSubscriber.Lock()
 	defer c.subStateSubscriber.Unlock()
 	c.subStateSubscriber.updateCh = updateCh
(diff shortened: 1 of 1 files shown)


lucab:
@dtnaylor these are the safety changes I previously mentioned, I'd be glad if you had a look.

/cc @dgonyeo for review

dbus/subscription.go
```diff
@@ -169,6 +169,13 @@ type SubStateUpdate struct {
 // is full, it attempts to write an error to errCh; if errCh is full, the error
 // passes silently.
 func (c *Conn) SetSubStateSubscriber(updateCh chan<- *SubStateUpdate, errCh chan<- error) {
+	if c == nil {
+		select {
+		case errCh <- errors.New("nil receiver"):
```
dgonyeo: I'm not familiar with this codebase, but is there a chance of a situation where writing to `errCh` will block indefinitely? Returning without propagating the error up in any way makes me nervous. Maybe log something at the very least in the `default` case?

lucab: `if errCh is full, the error passes silently` says the public docstring and other full errCh are ignored too. I understand your feeling though, and I'm trying to cleanup a bit the code. Should I go ahead and sprinkle `log.Printf` around?

dbus/subscription.go
```diff
@@ -227,10 +251,13 @@ func (c *Conn) shouldIgnore(path dbus.ObjectPath) bool {
 }
 
 func (c *Conn) updateIgnore(path dbus.ObjectPath, info map[string]interface{}) {
-	c.cleanIgnore()
```
dtnaylor: Why move this out of here?

(seems fine to me either way, but if anyone other than `sendSubStateUpdate` starts using the ignore functions, then they'll have to remember to call `cleanIgnore` themselves)

lucab: Because `updateIgnore` is only called after we a proper and non-ignored `info`, while `cleanIgnore` is generic bookkeeping that doesn't need any of it. I was just reading the code, and I felt it was out of place, so I turned it into a non-conditional defer.

You are right about the multiple callers. This code may need some more intrusive re-architecting once we get there, the whole ignore thing is quite hackish.

dtnaylor: Makes sense—sounds good.

lucab:
Rebased against master for conflicts. No further changes, retaining previous review.

PR #251, opened 2018-02-27
activation: add two functions to provide listeners with names

vincentbernat:
Since v227, systemd can name each listener. This is useful when you
have different socket units activating the same service: you can't
really know in which order the socket will be provided. Systemd adds
an environment variable named LISTEN_FDNAMES for this purpose.

We encapsulate the name in the File object created from the file
descriptor. Unfortunately, once this is turned into a net.Listener,
there doesn't seem to have a way to fetch this information
back. Therefore, two new functions are provided to be able to easily
get named sockets.

The same could be done to PacketConns in the future.

### DIFF
--- a/activation/listeners.go
+++ b/activation/listeners.go
@@ -37,6 +37,27 @@ func Listeners(unsetEnv bool) ([]net.Listener, error) {
 	return listeners, nil
 }
 
+// ListenersWithNames maps a listener name to a set of net.Listener instances.
+func ListenersWithNames(unsetEnv bool) (map[string][]net.Listener, error) {
+	files := Files(unsetEnv)
+	listeners := map[string][]net.Listener{}
+
+	for _, f := range files {
+		if pc, err := net.FileListener(f); err == nil {
+			current, ok := listeners[f.Name()]
+			if !ok {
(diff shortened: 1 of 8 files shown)


activation/listeners.go
```diff
@@ -37,6 +37,24 @@ func Listeners(unsetEnv bool) ([]net.Listener, error) {
 	return listeners, nil
 }
 
+// ListenersWithNames maps a listener name to a set of net.Listener instances.
+func ListenersWithNames(unsetEnv bool) (map[string][]net.Listener, error) {
+	files := Files(unsetEnv)
+	listeners := map[string][]net.Listener{}
+
+	for _, f := range files {
+		if pc, err := net.FileListener(f); err == nil {
```
lucab: I think this would benefit from the same conditional `Close` we are discussing at https://github.com/coreos/go-systemd/pull/250.

lucab:
For reference, this PR looks fine overall. I was just blocked trying to get a better view on https://github.com/coreos/go-systemd/pull/250.

vincentbernat:
I have also updated this PR. It should merge cleanly once #250 is merged.

PR #252, opened 2018-02-28
test: make dbus subscription tests more leniant

lucab:
This increases timeouts and adds optimistically-sized buffers to
communication channels, to reduce dbus tests flakiness.


lucab:
This eliminates flakes as seen in https://travis-ci.org/coreos/go-systemd/jobs/346828197 (from @vincentbernat's https://github.com/coreos/go-systemd/pull/250).

/cc @squeed for review

PR #254, opened 2018-03-03
sdjournal: add formating feature to JournalReader

### LINKED ISSUE #253: Feature request: Custom formatter for sdjournal reader
At the moment, `JournalReader` has a fixed message template for each entry. It would be really helpful if I would be able to alter the format.

I have thought of something like this:

```golang
config := sdjournal.JournalReaderConfig {
		Formatter: func(entry *JournalEntry) string {
                     return fmt.Sprintf("%s:%s",entry.Fields["MY_CUSTOM_FIELD"])
                },
	}
```

I can send you a patch if you are okay with this kind of feature.

Richterrettich:
this allows users of JournalReader to provide a callback function
that will format the output of each entry.

Fixes #253

### DIFF
--- a/sdjournal/read.go
+++ b/sdjournal/read.go
@@ -44,6 +44,11 @@ type JournalReaderConfig struct {
 	// If not empty, the journal instance will point to a journal residing
 	// in this directory. The supplied path may be relative or absolute.
 	Path string
+
+	// If not nil, Formatter will be used to translate the resulting entries
+	// into strings. If not set, the default format (timestamp and message field)
+	// will be used. If Formatter returns an error, Read will stop and return the error.
+	Formatter func(entry *JournalEntry) (string, error)
 }
 
 // JournalReader is an io.ReadCloser which provides a simple interface for iterating through the
@@ -51,12 +56,20 @@ type JournalReaderConfig struct {
 type JournalReader struct {
 	journal   *Journal
 	msgReader *strings.Reader
+	formatter func(entry *JournalEntry) (string, error)
 }
 
 // NewJournalReader creates a new JournalReader with configuration options that are similar to the
 // systemd journalctl tool's iteration and filtering features.
 func NewJournalReader(config JournalReaderConfig) (*JournalReader, error) {
-	r := &JournalReader{}
+	// use simpelMessageFormatter as default formatter.
+	if config.Formatter == nil {
+		config.Formatter = simpelMessageFormatter
+	}
+
+	r := &JournalReader{
+		formatter: config.Formatter,
+	}
 
 	// Open the journal
 	var err error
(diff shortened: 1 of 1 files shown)


sdjournal/read.go
```diff
@@ -44,19 +44,27 @@ type JournalReaderConfig struct {
 	// If not empty, the journal instance will point to a journal residing
 	// in this directory. The supplied path may be relative or absolute.
 	Path string
+
+	// If not nil, Formatter will be used to translate the resulting entries
+	// into strings. If not set, the default format (timestamp and message field)
+	// will be used.
+	Formatter func(entry *JournalEntry) string
```
lucab: I would prefer to make this more error-friendly by returning a `(string, error)` instead. As a consumer, would it be ok?

sdjournal/read.go
```diff
@@ -245,6 +253,13 @@ func (r *JournalReader) buildMessage() (string, error) {
 	var usec uint64
 	var err error
 
+	if r.formatter != nil {
+		entry, err := r.journal.GetEntry()
```
lucab: Given that we are introducing a formatting function, I would suggest to lift the entry fetching logic into the common path at `Read`, then change the `buildMessage` signature to match the formatter one, and just call `r.formatter(entry)` there.

lucab:
Thanks for your PR! I generally like the idea, I've just left a couple of comments to streamline the public interface and the integration point.

Richterrettich:
Thanks for the review. I will rework the patch and inlcude your suggestions.

Richterrettich:
ping @lucab , I think I have addressed all your suggestions. Feel free to take a look.

sdjournal/read.go
```diff
@@ -44,19 +44,33 @@ type JournalReaderConfig struct {
...
 	journal   *Journal
 	msgReader *strings.Reader
+	formatter func(entry *JournalEntry) (string, error)
 }
 
 // NewJournalReader creates a new JournalReader with configuration options that are similar to the
 // systemd journalctl tool's iteration and filtering features.
 func NewJournalReader(config JournalReaderConfig) (*JournalReader, error) {
-	r := &JournalReader{}
+
```
lucab: Spurious leading newline.

sdjournal/read.go
```diff
@@ -238,20 +257,12 @@ process:
...
-
-	if msg, err = r.journal.GetData("MESSAGE"); err != nil {
-		return "", err
-	}
-
-	if usec, err = r.journal.GetRealtimeUsec(); err != nil {
-		return "", err
-	}
+func simpelMessageFormatter(entry *JournalEntry) (string, error) {
+	msg := entry.Fields["MESSAGE"]
```
lucab: Please check for key presence:
```golang
msg, ok := entry.Fields["MESSAGE"]
if !ok { ... }
```

Richterrettich:
Yeah no problem. I somehow thougth that the "MESSAGE" field could never be empty in a journal entry. But a check can't hurt. I fixed both issues. Can you take a look again please?

PR #261, opened 2018-03-20
Linter warnings fixed for unit sources

muesli:
- Added missing godoc-style code comments
- Omitted redundant second value from range


lucab:
LGTM. Was this detected by some open-source checker tool that could be easily added to travis?

muesli:
You could try `gometalinter` with specific checks. I don't recommend running it with all linters though, esp. when cgo comes into play (and you want to keep constants in ALL_CAPS format for example).

PR #265, opened 2018-03-20
journal: Added missing error handling for write & unlink calls

muesli:
I don't think the first check can harm, the second one was more of a typo and the `err` assignment obviously missing.


squeed:
LGTM, thanks!

PR #270, opened 2018-03-22
daemon: Clean up copyright headers and documentation

### LINKED ISSUE #83: daemon/sdnotify.go probably violates Apache license
Hi go-systemd team!

Many thanks for all the work you do!

I'm packaging go-systemd for Ubuntu (as part of the juju project).
We do license checks for all the code we package as part of our packaging pipeline.
It seems to me that one file in your repo violates requirements of the Apache-2.0 license.
Could you please help me to understand if it's true or not and (if so) fix the issue in your repo?

File in question: https://github.com/coreos/go-systemd/blob/master/daemon/sdnotify.go

File states that it is 'forked from Docker project' which is correct.
Original file from Docker project can be found here:
https://github.com/docker/docker/blob/master/pkg/systemd/sd_notify.go
In terms of Apache-2.0 license forked version of the file is called Derivative Work.

Docker project is licensed under Apache-2.0 license (as well as go-systemd project).
This license allows you to redistribute Derivative Works only if you obey the following rules:

> (a) You must give any other recipients of the Work or Derivative Works a copy of this License;

While both projects are license under Apache-2.0 license, text is a bit different between the licenses.
Go-systemd license: https://github.com/coreos/go-systemd/blob/master/LICENSE
Docker license: https://github.com/docker/docker/blob/master/LICENSE
You may see that _at least_ formatting of paragraph 4 is different. It means that you need to include Docker's license into your project (maybe put it inside daemon/ folder?)
and put a reference to it from sdnotify.go file.

>  (b) You must cause any modified files to carry prominent notices stating that You changed the files;

While 'forked from Docker project' looks okay in the sense of this requirement I'd recommend you to specify which exact file you've modified (link to Docker's github repo may work).

> (c) You must retain, in the Source form of any Derivative Works that You distribute, all copyright, patent, trademark, and attribution notices <...>

Docker's sd_notify.go file doesn't contain explicit copyright notices but they are available in Docker's LICENSE file:
https://github.com/docker/docker/blob/master/LICENSE#L179
You may find there: 'Copyright 2013-2015 Docker, Inc.'

This copyright information needs to be included into daemon/sdnotify.go.
You may add your own copyright notice (for your changes) but you shouldn't remove Docker's copyright.

> (d) If the Work includes a "NOTICE" text file <...> then any Derivative Works <...> must include a readable copy of the attribution notices contained within such NOTICE file <...>

Docker project has such NOTICE file: https://github.com/docker/docker/blob/master/NOTICE
This file contains some valuable information and I'd recomment you to put it into daemon/ folder as well.
It may be linked from daemon/sd_notify.go as well as the LICENSE file.

Thanks A LOT for helping!
Oleg

muesli:
- Added copyright 2015-2018 CoreOS, Inc.
- Proper godoc package comment
- Got rid of named (but unused) return values

Fixes #83.

### DIFF
--- a/daemon/sdnotify.go
+++ b/daemon/sdnotify.go
@@ -1,4 +1,5 @@
 // Copyright 2014 Docker, Inc.
+// Copyright 2015-2018 CoreOS, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -13,7 +14,11 @@
 // limitations under the License.
 //
 
-// Code forked from Docker project
+// Package daemon provides a Go implementation of the sd_notify protocol.
+// It can be used to inform systemd of service start-up completion, watchdog
+// events, and other status changes.
+//
+// https://www.freedesktop.org/software/systemd/man/sd_notify.html#Description
 package daemon
 
 import (
@@ -29,7 +34,7 @@ import (
 // (false, nil) - notification not supported (i.e. NOTIFY_SOCKET is unset)
 // (false, err) - notification supported, but failure happened (e.g. error connecting to NOTIFY_SOCKET or while sending data)
 // (true, nil) - notification supported, data has been sent
-func SdNotify(unsetEnvironment bool, state string) (sent bool, err error) {
+func SdNotify(unsetEnvironment bool, state string) (bool, error) {
 	socketAddr := &net.UnixAddr{
 		Name: os.Getenv("NOTIFY_SOCKET"),
 		Net:  "unixgram",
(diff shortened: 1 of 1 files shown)


daemon/sdnotify.go
```diff
@@ -41,10 +46,10 @@ func SdNotify(unsetEnvironment bool, state string) (sent bool, err error) {
 	}
 
 	if unsetEnvironment {
-		err = os.Unsetenv("NOTIFY_SOCKET")
-	}
-	if err != nil {
-		return false, err
+		err := os.Unsetenv("NOTIFY_SOCKET")
+		if err != nil {
```
lucab: You can merge this and the above in a single statement.

daemon/sdnotify.go
```diff
@@ -55,9 +60,5 @@ func SdNotify(unsetEnvironment bool, state string) (sent bool, err error) {
 	defer conn.Close()
 
 	_, err = conn.Write([]byte(state))
-	// Error sending the message
-	if err != nil {
-		return false, err
-	}
-	return true, nil
+	return err == nil, err
```
lucab: I would keep logic outside of return statement.
Even if I agree the original form may be overly verbose, I think it is the expected idiomatic form.

muesli:
@lucab Done and squashed.

PR #271, opened 2018-03-22
daemon: add convenience constants for SdNotify

muesli:
This provides convenient little helper methods for the common service notifications, while hiding the "protocol" details from the API caller.

- SdNotifyReady to signal readiness
- SdNotifyStopping to signal beginning of shutdown
- SdNotifyReloading to signal configuration reloading
- SdNotifyWatchdog to update the watchdog timestamp

### DIFF
--- a/daemon/sdnotify.go
+++ b/daemon/sdnotify.go
@@ -21,6 +21,25 @@ import (
 	"os"
 )
 
+const (
+	// SdNotifyReady tells the service manager that service startup is finished
+	// or the service finished loading its configuration.
+	SdNotifyReady = "READY=1"
+
+	// SdNotifyStopping tells the service manager that the service is beginning
+	// its shutdown.
+	SdNotifyStopping = "STOPPING=1"
+
+	// SdNotifyReloading tells the service manager that this service is
+	// reloading its configuration. Note that you must call SdNotifyReady when
+	// it completed reloading.
+	SdNotifyReloading = "RELOADING=1"
+
+	// SdNotifyWatchdog tells the service manager to update the watchdog
+	// timestamp for the service.
+	SdNotifyWatchdog = "WATCHDOG=1"
+)
+
 // SdNotify sends a message to the init daemon. It is common to ignore the error.
 // If `unsetEnvironment` is true, the environment variable `NOTIFY_SOCKET`
 // will be unconditionally unset.
--- a/daemon/watchdog.go
+++ b/daemon/watchdog.go
@@ -21,10 +21,11 @@ import (
 	"time"
 )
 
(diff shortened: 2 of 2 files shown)


lucab:
I'm in general not a big fan of these one-line wrappers. I'd still gladly take a commit to fix the docstring, adding the missing boolean parameter.

muesli:
@lucab Understood, even though I'd like to point out that they are here for more than just convenience. They hide away the protocol specific details most go-systemd users wouldn't and shouldn't be familiar with. As such I'd hope you would reconsider their purpose once more.

lucab:
Perhaps we can provide pre-baked constants for those? In general we are not trying to hide protocol details away, and we often reference manpages where they are abundantly detailed. We just provide thin-wrappers to consume them in a safer idiomatic way.

Pinging @squeed and @euank for some third-party feedback.

muesli:
Just to make it perfectly clear: I can absolutely understand your reasoning and that's an acceptable design decision by any means.

However I must admit that I think that go-systemd already does provide a bunch of convenient wrappers (all the dbus handling comes to mind, for example), which makes it feel like a natural fit for those abstractions, too.

No hard feelings, though, I completely trust in you making the right decision for this library and its user base!

euank:
I do think it's aligned with the goals of this package to make those sdnotify calls easier to write. Having consumers write magic strings, especially given the same few strings will be by and far the most commonly used ones, seems needlessly fragile.

I don't think that the functions here are actually much more helpful than consts though. As is, the functions aren't providing much abstraction. Both `daemon.SdNotify(false, daemon.NotifyReady)` and `daemon.SdNotifyReady()` encode basically the same information (other than the meaning of the first parameter).

Based on the above, I'm leaning towards defining consts being a better approach to making it easier for consumers of this package while avoiding a too-thin layer of abstraction.

@lucab @muesli Does providing consts for those common state strings make sense to both of you?

muesli:
@euank Sounds absolutely reasonable to me! Happy to update the PR if that's the consensus.

lucab:
Yup, let's just add those constants then.

muesli:
@lucab @euank Squshed and pushed the requested changes. I moved all the consts into sdnotify.go to keep them in one place. Thanks for the reviews!

lucab:
LGTM, thanks everybody!

GwynethLlewelyn:
I know, I'm bumping an ancient thread, but it would be nice to add a few more to the list, e.g. see https://www.freedesktop.org/software/systemd/man/sd_notify.html for a long list... some are quite useful, such as sending back a status, for a particularly long-running process.

PR #273, opened 2018-03-27
sdjournal: Fixed race conditions in JournalReader.Follow

muesli:
JournalReader.Follow needs to wait for r.Journal.Wait to return, otherwise
we're opening the door for several race conditions. The JournalReader could
either be re-used or closed while Wait is still running, which seems to
result in undefined behavior. Commonly Wait will return a non-existing,
negative status code in such cases.

Since Wait might return after we stopped reading from waitCh, it's now a
buffered channel of size 1, so it'll never dead-lock in such situations.

Read errors are now correctly propagated to the caller, since err isn't
shadowing the named return value any longer.

We also abort when receiving unknown status codes from Wait now, to prevent
infinitely looping over it.

### DIFF
--- a/sdjournal/read.go
+++ b/sdjournal/read.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"log"
 	"strings"
+	"sync"
 	"time"
 )
 
@@ -195,20 +196,20 @@ func (r *JournalReader) Rewind() error {
 
 // Follow synchronously follows the JournalReader, writing each new journal entry to writer. The
 // follow will continue until a single time.Time is received on the until channel.
-func (r *JournalReader) Follow(until <-chan time.Time, writer io.Writer) (err error) {
+func (r *JournalReader) Follow(until <-chan time.Time, writer io.Writer) error {
 
 	// Process journal entries and events. Entries are flushed until the tail or
 	// timeout is reached, and then we wait for new events or the timeout.
 	var msg = make([]byte, 64*1<<(10))
-	var waitCh = make(chan int)
-	var waitStop = make(chan bool)
-	defer close(waitStop)
+	var waitCh = make(chan int, 1)
+	var waitGroup sync.WaitGroup
+	defer waitGroup.Wait()
 
 process:
 	for {
 		c, err := r.Read(msg)
 		if err != nil && err != io.EOF {
-			break process
+			return err
 		}
 
 		select {
@@ -218,7 +219,7 @@ process:
 		}
 		if c > 0 {
 			if _, err = writer.Write(msg[:c]); err != nil {
-				break process
+				return err
 			}
 			continue process
 		}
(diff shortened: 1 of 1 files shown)


muesli:
An example triggering this undefined behavior or infinitely looping over the reader, you could try https://github.com/muesli/service-tools (in particular service-monitor). If there's anything else I can comment on to help clarify the underlying issue, please let me know.

sdjournal/read.go
```diff
@@ -244,13 +246,14 @@ process:
 				case SD_JOURNAL_APPEND, SD_JOURNAL_INVALIDATE:
 					continue process
 				default:
-					log.Printf("Received unknown event: %d\n", e)
+					err = fmt.Errorf("received unknown event: %d", e)
+					break process
```
lucab: It looks like we are using a mix of `break process` and `return err`. As this is removing the named return (:+1:) I think it would make sense to also get rid of the `break`s to make this a bit more readable. What do you think?

sdjournal/read.go
```diff
@@ -244,13 +246,14 @@ process:
 				case SD_JOURNAL_APPEND, SD_JOURNAL_INVALIDATE:
 					continue process
 				default:
-					log.Printf("Received unknown event: %d\n", e)
+					err = fmt.Errorf("received unknown event: %d", e)
```
lucab: I think that the infinite loop here was on purpose to make this a bit more resilient against unknown/future states. The `until` parameter already takes care of bounding the loop. Unless you have more specific cases I am not aware of, I'd leave this as a log statement.

lucab:
Your PR was clear enough, thanks! Some personal issues delayed my reviewed on this, sorry. I just added  my comments now.

sdjournal/read.go
muesli: Thanks, you're right, the code can be further simplified, now! I used to not `defer` the `Wait` and call it manually before the return, but that makes it even nicer. Changing!

sdjournal/read.go
muesli: My personal use-case is a log-viewer that follows the journal until the user exits the program. Strictly from my personal experience, this imposes two issues:

- Once I triggered the race condition (and hence an invalid reader-state), this code kept infinitely looping over the same invalid state. Granted, this PR would prevent me from triggering the race condition in the future, but:

- While this is obviously an unexpected error, it would still be nice for the package's user to detect an invalid / bad state and act accordingly, e.g. present the error to the user (in a friendlier way than spamming the terminal) or cancelling the `JournalReader` via the `until` channel.

As a rule of thumb: this being a package and not an application, I would prefer it to forward all its errors (that can't be recovered from internally) to its users and not `panic` / `log` if possible.

muesli:
@lucab No worries, completely understood! I was just wondering if I did a bad job at explaining the situation :)

sdjournal/read.go
lucab: I understand your stance, and your comments are on spot, I'm just trying to find a good balance here. According to `sd_journal_wait` manpage, on failures it will return negative error codes. Can we adjust this default case to check for `e < 0` and return only in that case? I'd still keep the logging for unexpected-but-successful codes.

muesli: I'm fine with that. Will update the PR today!

muesli:
@lucab Ready for another review!

PR #276, opened 2018-05-07
activation: remove unsafe support for fd re-use

lucab:
Package `activation` was previously offering an option to leave
environment variables untouched, in order to re-use fd at a future
point.
However, re-using fd is unsafe and results in racing against the GC
due to the way how `os.File` is implemented. Moreover, the Go runtime
does not (nor plan to) provide a safe interface for proper temporary
borrowing of existing fd.

Thus, this removes the `unsetEnv` option from all listeners (forcing
to always unset), and only keeps it where ownership can be transferred
back to the caller.

Ref: https://github.com/coreos/go-systemd/pull/250#issuecomment-369625901
Ref: https://github.com/golang/go/issues/24215

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ env:
     - GOPATH=/opt
     - BUILD_DIR=/opt/src/github.com/coreos/go-systemd
   matrix:
-    - DOCKER_BASE=ubuntu:16.04
+    - DOCKER_BASE=ubuntu:18.04
     - DOCKER_BASE=debian:stretch
 
 before_install:
--- a/activation/files.go
+++ b/activation/files.go
@@ -22,11 +22,17 @@ import (
 	"syscall"
 )
 
-// based on: https://gist.github.com/alberts/4640792
 const (
+	// listenFdsStart corresponds to `SD_LISTEN_FDS_START`.
 	listenFdsStart = 3
 )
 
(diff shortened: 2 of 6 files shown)


lucab:
/cc @squeed @vincentbernat for review

This is a followup to https://github.com/coreos/go-systemd/pull/250#issuecomment-374564296. It contains a breaking change for the `activation` package API.

lucab:
I also had to bump the ubuntu image we use for tests, as 16.04 comes with an ancient go compiler, and as one of our dependencies (godbus) started using `context`, and as we don't vendor in here (as this is a library), and as `go get` does not support checking out a specific tag.

vincentbernat:
OK for me.

vincentbernat:
`Files()` is always called with `true`. It could also loose its argument.

lucab:
@vincentbernat yes, we only use it that way. But it is part of the public API and its usage is actually safe, so I plan to keep it as is.

squeed:
Looks fine to me! Do you want to add something to the docstring like "you should pass unsetEnv as true unless you know what you're doing"?

lucab:
@squeed good idea. I added a couple of lines to describe fag rationale and typical usage. Everything else kept untouched, PTAL.
I'd plan to tag `v17` right after this.

Issue #275, opened 2018-04-21
journald: support mocking Send for external consumers


toravir:
Is there a way to write go tests to test journal.Send() calls ?? So that i can know for sure all the pieces of data sent are indeed sent to the journald correctly.

toravir:
sdjournal tests writing and reading. But in my package i am just using journald to write. I cannot use sdjournal to read back - since sdjournal requires the underlying system to have libsystemd-dev. Is there some mock mechanism to check if the data is sent correctly over the unix Socket ??

lucab:
At the moment a consumer cannot mock or hijack the socket for `journal.Send`. While I am not sure this is something we actually want to expose, I think it may be feasible after the rework described in https://github.com/coreos/go-systemd/pull/218#issuecomment-274490875 is done.

Why do you want to externally test `journal.Send` behavior? This should pertain to library tests, without exposing all internals knobs to consumers.
Can you just hijack `/run/systemd/journal/socket` in your environment?

toravir:
A library that uses journal.Send(), needs to write a test - like in https://github.com/rs/zerolog/blob/master/journald/journald_test.go, it can verify the data being given to the library is sent correctly to the systemd.

I want to write go-tests - so i am looking for some mock that is process specific (so that the test can run in an env with and without systemd). It is just like mock "net/http/httptest" package. I can code up and send a PR if you are willing to consider a mock ??

lucab:
If you want to check that the data has been properly recorded in journald then the way to go is indeed the `sdjournal` package, which requires a running systemd-journald + libsystemd. If you aren't in such environment and don't actually care about journald behavior, then just place your own receiver behind a unix socket at `/run/systemd/journal/socket`.

In general no, I'm not eager to expose testing/mocking interface through this library. The further we can go in here is making the socket endpoint configurable, but even that currently requires ground rework (see link above).

lucab:
@toravir FYI https://github.com/coreos/go-systemd/pull/279/files#diff-985ded8023b21268d0d5cf7f29781d2eR55.

Issue #233, opened 2017-05-26
Journal logging doesn't reconnect if journald gets restarted


krobertson:
We encountered this issue today twice on a production system. We haven't yet identified the root cause, however this appeared to be an ancillary issue.

The host was running out of memory, likely from journald. It attempted to compress and rotate logs, but was unable to allocate the memory. Journald died but was restarted. However' we're also using the docker journald log driver along with journald's syslog transport. After the restart, all docker containers failed to continue logging with `EPIPE` errors to stdout/stderr.

In looking at the `journal` code, it connects to journald on init, however doesn't have any error handling where it might reconnect if journald was restarted.

The log output we had was:

```
May 26th 2017, 05:34:03.204    System journal (/var/log/journal/) is 1.7G, max 2.6G, 931.8M free.
May 26th 2017, 05:34:04.220    Failed to initialize XZ encoder: code 5
May 26th 2017, 05:34:04.231    systemd-journald.service: Main process exited, code=dumped, status=6/ABRT
May 26th 2017, 05:34:04.232    Failed to compress (unnamed temporary file): Invalid argument
May 26th 2017, 05:34:04.242    Detected coredump of the journal daemon or PID 1, diverted to /var/lib/systemd/coredump/core.systemd-journal.0.9b634f7d87464833a67ca9124f25ab86.14979.1495802022000000.
May 26th 2017, 05:34:04.242    systemd-journald.service: Unit entered failed state.
May 26th 2017, 05:34:04.245    systemd-journald.service: Service has no hold-off time, scheduling restart.
May 26th 2017, 05:34:04.245    systemd-journald.service: Failed with result 'core-dump'.
May 26th 2017, 05:34:04.259    Stopped Flush Journal to Persistent Storage.
May 26th 2017, 05:34:04.259    Stopped Journal Service.
May 26th 2017, 05:34:04.259    Stopping Flush Journal to Persistent Storage...
May 26th 2017, 05:34:04.268    Starting Journal Service...
```

After that, we only get huge error spikes to New Relic with `EPIPE`.

lucab:
See https://github.com/coreos/go-systemd/pull/218 for reference.

ssgreg:
Guys, please check the implementation: https://github.com/ssgreg/journald

PR #279, opened 2018-05-28
journal: use a connection-less socket

lucab:
This gets rid of the connection ownership problem by using a non-connected Unix datagram socket and an explicit destination to send each message. It does not impact the public API, but it requires a minimum-go version bump due to https://go-review.googlesource.com/c/go/+/45872 (golang 1.10+).

Fixes https://github.com/coreos/go-systemd/issues/5
Fixes https://github.com/coreos/go-systemd/issues/233
Fixes https://github.com/coreos/go-systemd/issues/275

---

Note: as 1.11 got released, all upstream supported versions of the Go runtime have a fixed version of `WriteMsgUnix` that this PR uses. However, this means that the `journal` package won't work on older/buggy runtimes (<= 1.9).

### DIFF
--- a/journal/journal.go
+++ b/journal/journal.go
@@ -33,7 +33,10 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
+	"sync/atomic"
 	"syscall"
+	"unsafe"
 )
 
 // Priority of a journal message
@@ -50,19 +53,35 @@ const (
 	PriDebug
 )
 
-var conn net.Conn
+var (
+	// This can be overridden at build-time:
+	// https://github.com/golang/go/wiki/GcToolchainTricks#including-build-information-in-the-executable
+	journalSocket = "/run/systemd/journal/socket"
+
+	// unixConnPtr atomically holds the local unconnected Unix-domain socket.
+	// Concrete safe pointer type: *net.UnixConn
+	unixConnPtr unsafe.Pointer
+	// onceConn ensures that unixConnPtr is initialized exactly once.
+	onceConn sync.Once
+)
 
 func init() {
-	var err error
-	conn, err = net.Dial("unixgram", "/run/systemd/journal/socket")
-	if err != nil {
-		conn = nil
-	}
+	onceConn.Do(initConn)
 }
 
-// Enabled returns true if the local systemd journal is available for logging
+// Enabled checks whether the local systemd journal is available for logging.
 func Enabled() bool {
-	return conn != nil
+	onceConn.Do(initConn)
+
+	if (*net.UnixConn)(atomic.LoadPointer(&unixConnPtr)) == nil {
+		return false
+	}
+
+	if _, err := net.Dial("unixgram", journalSocket); err != nil {
+		return false
+	}
+
+	return true
 }
 
 // Send a message to the local systemd journal. vars is a map of journald
@@ -73,8 +92,14 @@ func Enabled() bool {
 // (http://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html)
 // for more details.  vars may be nil.
 func Send(message string, priority Priority, vars map[string]string) error {
+	conn := (*net.UnixConn)(atomic.LoadPointer(&unixConnPtr))
 	if conn == nil {
-		return journalError("could not connect to journald socket")
+		return errors.New("could not initialize socket to journald")
+	}
+
+	socketAddr := &net.UnixAddr{
+		Name: journalSocket,
+		Net:  "unixgram",
 	}
 
 	data := new(bytes.Buffer)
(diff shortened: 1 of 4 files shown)


journal/journal.go
```diff
@@ -74,7 +90,12 @@ func Enabled() bool {
 // for more details.  vars may be nil.
 func Send(message string, priority Priority, vars map[string]string) error {
 	if conn == nil {
-		return journalError("could not connect to journald socket")
+		return errors.New("could not initialize socket")
+	}
+
+	socketAddr, err := net.ResolveUnixAddr("unixgram", journalSocket)
```
toravir: Is it needed to ResolveUnixAddr on every Send() call ? Since journalSocket is not going to change once the library is inited.

lucab: Indeed, there is no need for `ResolveUnixAddr`, we can just create our `UnixAddr` struct manually here instead.

lucab:
@cachedout Yes, I think it is fine to merge it now, but I need to refresh it and grab a reviewer. Do you have any urgent need for this?

cachedout:
@lucab I suspect this is a root cause for this bug: https://github.com/elastic/beats/issues/9533

I don't think it's a huge rush, personally :)

Thanks so much for the speedy response!

lucab:
@cachedout I refreshed this and gave a few runs in CI, without seeing any flakes. If you can get some feedback from the field before I merge this, I would be happy to hear that.

@s-urbaniak @ashcrow I'd summon you as reviewers for this beast from the past. If you have time. Pretty please with sugar.

journal/journal.go
```diff
@@ -50,19 +50,35 @@ const (
...
 	if err != nil {
 		conn = nil
+		return
 	}
 }
 
 // Enabled returns true if the local systemd journal is available for logging
 func Enabled() bool {
-	return conn != nil
+	if conn == nil {
```
s-urbaniak: admittedly, this is unrelated to this PR, but I wonder if `init()` has any happens-before guarantees in the runtime.

If `Enabled()` is being invoked in a different goroutine other than the main goroutine which invoked `init()` this would be an unsynchronized read, potentially returning any arbitrary value.

cachedout:
@lucab I'll see what I can do to verify this fix ASAP. Thanks!

journal/journal.go
s-urbaniak: nevermind, `init()` _is_ part of https://golang.org/ref/spec#Package_initialization, so happens-before should apply here :+1:

journal/journal.go
```diff
@@ -84,31 +104,27 @@ func Send(message string, priority Priority, vars map[string]string) error {
...
+	if err != nil {
+		if !isSocketSpaceError(err) {
+			return err
+		}
+
+		// Large log entry, send it via tempfile and ancillary-fd.
 		file, err := tempFd()
 		if err != nil {
-			return journalError(err.Error())
+			return err
```
s-urbaniak: nit: some error wrapping would be nice here to have context at the caller's side.

journal/journal.go
```diff
@@ -84,31 +104,27 @@ func Send(message string, priority Priority, vars map[string]string) error {
 		appendVariable(data, k, v)
 	}
 
-	_, err := io.Copy(conn, data)
-	if err != nil && isSocketSpaceError(err) {
+	_, _, err := conn.WriteMsgUnix(data.Bytes(), nil, socketAddr)
+	if err != nil {
```
s-urbaniak: very small stylistic nit to remove indentation:
```
if err == nil {
  return nil
}

// unindented code
```

journal/journal.go
lucab: That's a fair concern though. While goroutines can't race here, other packages' `init` functions calling `journal.Enabled` could be scheduled before our `init`. If that is a concern, possible solutions would be:
 * avoid global variables (which is tempting, but moves socket creation in the hot path of log emitting)
 * wrap `conn` initialization in a `sync.Once` and use it here too

journal/journal.go
```diff
@@ -50,19 +50,35 @@ const (
...
 }
 
 // Enabled returns true if the local systemd journal is available for logging
 func Enabled() bool {
-	return conn != nil
+	if conn == nil {
+		return false
+	}
+	_, err := net.ResolveUnixAddr("unixgram", journalSocket)
+	if err == nil {
```
lucab: Note to self: this is a tautological condition due to `ResolveUnixAddr` implementation, and can be removed.

journal/journal.go
s-urbaniak: `sync.Once` sounds like a good solution :+1:

blufor:
@bdwyertech @toravir @s-urbaniak Hi guys, can I be of a help somehow? We're waiting for this for this to be merged so that we can use `journalbeat` to ship logs from journald on all of our infrastructure. I'm not sure, what's blocking this now

ashcrow:
Sorry for the late review, but LGTM

iakovmarkov:
Great news! @lucab can you merge this, please?

lucab:
I was hopefully waiting for @cachedout to verify that it fixes beats issues (or at least that it doesn't introduce new regressions). Did/will anyboy confirm that?

In any case, I will do another final-checking pass on this (as soon as I have free timeslot) and then merge if nothing new shows up.

kvaps:
From my point of view this fix is working pretty well.
Now I have really stable flow from my journalbeats after applying the patch.
Good job @lucab, thank you!

kvch:
@lucab Thank you for fixing this issue! This definitely helped us a lot.

I would like to update the vendored code in `elastic/beats`. Are you planning to release a new version with this fix?

lucab:
@kvch tagged `v19` (which includes this PR): https://github.com/coreos/go-systemd/releases/tag/v19

kvch:
Awesome. Thank you!

PR #285, opened 2018-08-28
journal: fix validation of key names

### LINKED ISSUE #284: validVarName doesn't fail variables starting with _
Variables starting with an underscore are not allowed by systemd. The code tries to test for it but there's a mistake with a missing parentheses around the or clauses.

GraveRaven:
the validVarName function tries to test for variables starting with an
underscore but there's a missing set of parentheses.

Fixes #284

### DIFF
--- a/journal/journal.go
+++ b/journal/journal.go
@@ -119,8 +119,8 @@ func Print(priority Priority, format string, a ...interface{}) error {
 }
 
 func appendVariable(w io.Writer, name, value string) {
-	if !validVarName(name) {
-		journalError("variable name contains invalid character, ignoring")
+	if err := validVarName(name); err != nil {
+		journalError(err.Error())
 	}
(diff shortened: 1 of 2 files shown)


lucab:
@GraveRaven indeed, good catch! I can merge this as-is, but if you are up for some minor rework, I see that `validVarName` may benefit from a few other changes:
 * return an `error` instead of a bool, describing *why* the key name is invalid
 * check for the empty string, and fail validation in that case (with a testcase covering it)
 * move the internal comment as a normal docstring
 * link to the reference URL for key name requirements, i.e. https://www.freedesktop.org/software/systemd/man/sd_journal_print.html

Would you be ok tackling those here?

GraveRaven:
Yeah sure

lucab:
LGTM, @GraveRaven thanks for fixing and cleaning this up!

PR #288, opened 2018-09-30
machine1: add methods ListMachines, ListImages, GetMachineAddresses

dongsupark:
Add new DBUS methods exposed already by systemd.

* `ListMachines`: lists all active machines
* `ListImages`: lists all images under `/var/lib/machines`.
* `GetMachineAddresses`, which gets a list of IP addresses of the given machine name.

See also https://github.com/kinvolk/kube-spawn/issues/241#issuecomment-425708704

### DIFF
--- a/machine1/dbus.go
+++ b/machine1/dbus.go
@@ -37,6 +37,14 @@ type Conn struct {
 	object dbus.BusObject
 }
 
+// MachineStatus is a set of necessary info for each machine
+type MachineStatus struct {
+	Name    string          // The primary machine name as string
+	Class   string          // The machine class as string
+	Service string          // The machine service as string
+	JobPath dbus.ObjectPath // The job object path
+}
+
 // New() establishes a connection to the system bus and authenticates.
 func New() (*Conn, error) {
 	c := new(Conn)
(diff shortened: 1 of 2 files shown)


lucab:
Thanks a lot, looks very good on the surface! I'll try to find some time in the next few days to double-check the D-Bus side and for the final review.

machine1/dbus.go
```diff
@@ -191,3 +202,59 @@ func (c *Conn) ListMachines() ([]MachineStatus, error) {
 
 	return machs, nil
 }
+
+func imageFromInterfaces(image []interface{}) (*ImageStatus, error) {
+	if len(image) < 6 {
```
lucab: I think this should be `< 7`, right? Otherwise we may be off by one on the last `jobpath` (if I am not miscounting).

machine1/dbus_test.go
```diff
@@ -58,6 +60,17 @@ func mustCreateTestProcess(machineName string) (pid int) {
 	return int(mainPID)
 }
 
+func mustCreateTestImage(imagePath string) error {
```
lucab: I think this is small enough that it can be simply inlined into `TestImages` (which is the only consumer). Also the `defer os.Remove` should in-between `Create` and `Truncate` to avoid potentially leaking empty files.

lucab:
@dongsupark two minor things to change, looks good otherwise (DBus side too). Those methods have been around since many years now, so I think they don't need any additional version-compatibility note in their docstrings.

dongsupark:
@lucab Fixed two minor things. Thanks for the review!

lucab:
@dongsupark I plan to release this with `v18` once I've gone through the other things in the milestone: https://github.com/coreos/go-systemd/milestone/3. If you have other things you want to land before that (portable1?) feel free to open a placeholder ticket even if you don't have a PR yet, so I'll remember to wait for that.

dongsupark:
@lucab Right.
I filed a PR https://github.com/coreos/go-systemd/issues/291 for the milestone.

PR #292, opened 2018-10-12
docs: Add more detail to Conn.ListUnits-doc

### LINKED ISSUE #289: dbus: ListUnits does not behave as documented
[Conn.ListUnits](https://godoc.org/github.com/coreos/go-systemd/dbus#Conn.ListUnits) currently says:
>  ListUnits returns an array with all currently loaded units.

However there are been reports from the field that its behavior is somewhat puzzling and some loaded units are missing from the returned array. In particular, @SuperQ noticed that Prometheus node_exporter is encountering https://github.com/prometheus/node_exporter/issues/1082 in a consistent way across multiple OSes and systemd versions.

This code and documentation has been there for a long time, so it could be either a case of wrong documentation or behavioral change that slipped in. It would be good to read through systemd server-side implementation and then update go-systemd logic and/or documentation.

miono:
Cleared up some caveats about what units are returned from systemds
ListUnits-method.

Fixes #289

### DIFF
--- a/dbus/methods.go
+++ b/dbus/methods.go
@@ -291,6 +291,8 @@ func (c *Conn) listUnitsInternal(f storeFunc) ([]UnitStatus, error) {
 // ListUnits returns an array with all currently loaded units. Note that
 // units may be known by multiple names at the same time, and hence there might
 // be more unit names loaded than actual units behind them.
+// Also note that a unit is only loaded if it is active and/or enabled.
+// Units that are both disabled and inactive will thus not be returned.
 func (c *Conn) ListUnits() ([]UnitStatus, error) {
 	return c.listUnitsInternal(c.sysobj.Call("org.freedesktop.systemd1.Manager.ListUnits", 0).Store)
 }


dbus/methods.go
```diff
@@ -291,6 +291,8 @@ func (c *Conn) listUnitsInternal(f storeFunc) ([]UnitStatus, error) {
 // ListUnits returns an array with all currently loaded units. Note that
 // units may be known by multiple names at the same time, and hence there might
 // be more unit names loaded than actual units behind them.
+// Also note that a unit is only loaded if it is active and/or enabled.
+// Disabled and inactive units will thus not be returned.
```
lucab: My understanding it that this last line should be "Units that are both disabled and inactive...", right? That is, this is a strong logical AND.

lucab:
LGTM. You can squash those two commits together, and I'll merge.

miono:
Done.

lucab:
@miono merging, thanks again for cross-referencing and clarifying the docs!

PR #293, opened 2018-10-23
fix dbus api usage

h8liu:


lucab:
Yup. We stick to "Go libraries don't vendor", but we communicate dependencies/versions to consumers via the [dep manifest](https://github.com/coreos/go-systemd/blob/c6f51f82210d9608a835763225a5e2a3c87583c6/Gopkg.toml).

I'll be happy to take a similar PR once a new go-dbus release is tagged. That would also imply raising the minimum go-toolchain version to 1.7 though.

purpleidea:
@h8liu btw, sorry to re-write this patch. I didn't realize there was a dupe until I had sent mine. All merged now anyways. Woo!

PR #294, opened 2018-10-26
dbus: Fix dbus.SystemBus* signatures

trapperhoney:
The signatures for these function recently change to include an optional
options structure in
https://github.com/godbus/dbus/commit/0c07171c698a59782f0f12c1cea802d753e15794

Update references appropriately.

### DIFF
--- a/dbus/dbus.go
+++ b/dbus/dbus.go
@@ -143,7 +143,7 @@ func NewUserConnection() (*Conn, error) {
 func NewSystemdConnection() (*Conn, error) {
 	return NewConnection(func() (*dbus.Conn, error) {
 		// We skip Hello when talking directly to systemd.
-		return dbusAuthConnection(func() (*dbus.Conn, error) {
+		return dbusAuthConnection(func(opts ...dbus.ConnOption) (*dbus.Conn, error) {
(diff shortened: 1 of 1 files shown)


trapperhoney:
Thanks, I guess that explains why the build checks failed...  However, this breaks all of the packages that depend on this one (which is how I ended up here) since the dbus code is not included in a vendor directory of this project...  Maybe I'm missing some feature of 'go get' that is capable of parsing .toml files?

lucab:
@trapperhoney you are not missing that. However, you should instead pin and vendor dependencies in whatever application is consuming this. There are a few dependency managers to do that so you can pick the one you prefer. The TOML manifest here is in the [dep](https://github.com/golang/dep) format.

lucab:
duplicate of https://github.com/coreos/go-systemd/pull/296

go-dbus 5.0 got released, and the above PR fixed this same issue. Closing.

purpleidea:
@trapperhoney , sorry to re-write this patch. I didn't realize there was a dupe until I had sent mine. All merged now anyways. Woo!

trapperhoney:
Thank you both!  Mine wasn't the first patch either ;)

Issue #295, opened 2018-10-30
Someone broke the builds!


purpleidea:
```
$ go get -u github.com/coreos/go-systemd/dbus
# github.com/coreos/go-systemd/dbus
../gopath/src/github.com/coreos/go-systemd/dbus/dbus.go:127:33: cannot use dbus.SystemBusPrivate (type func(...dbus.ConnOption) (*dbus.Conn, error)) as type func() (*dbus.Conn, error) in argument to dbusAuthHelloConnection
../gopath/src/github.com/coreos/go-systemd/dbus/dbus.go:136:33: cannot use dbus.SessionBusPrivate (type func(...dbus.ConnOption) (*dbus.Conn, error)) as type func() (*dbus.Conn, error) in argument to dbusAuthHelloConnection
```

I don't know exactly when this started because I've been AFK a bit, but someone broke the builds! :(

lucab:
https://github.com/coreos/go-systemd/pull/293#issuecomment-432563677

lucab:
Closed by https://github.com/coreos/go-systemd/pull/296.

PR #296, opened 2018-10-30
dbus: Fix API break from godbus

purpleidea:
Something in godbus seems to have broken API, thus breaking builds
including `go get`. This patches fixes those changes. I don't know any
background about this change, so if anyone has information, I'd sure
like to know! Thanks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,9 @@ services:
 
 language: go
 go:
-  - "1.5.x"
+  - "1.7.x"
   - "1.10.x"
+  - "1.11.x"
 go_import_path: github.com/coreos/go-systemd
 
 env:
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -4,7 +4,7 @@
 
 [[constraint]]
   name = "github.com/godbus/dbus"
-  version = "4.1.0"
+  version = "5.0"
 
 [prune]
   go-tests = true
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Build Status](https://travis-ci.org/coreos/go-systemd.png?branch=master)](https://travis-ci.org/coreos/go-systemd)
 [![godoc](https://godoc.org/github.com/coreos/go-systemd?status.svg)](http://godoc.org/github.com/coreos/go-systemd)
-![minimum golang 1.5](https://img.shields.io/badge/golang-1.5%2B-orange.svg)
+![minimum golang 1.7](https://img.shields.io/badge/golang-1.7%2B-orange.svg)
 
 
 Go bindings to systemd. The project has several packages:
--- a/dbus/dbus.go
+++ b/dbus/dbus.go
@@ -143,7 +143,7 @@ func NewUserConnection() (*Conn, error) {
 func NewSystemdConnection() (*Conn, error) {
 	return NewConnection(func() (*dbus.Conn, error) {
 		// We skip Hello when talking directly to systemd.
-		return dbusAuthConnection(func() (*dbus.Conn, error) {
+		return dbusAuthConnection(func(opts ...dbus.ConnOption) (*dbus.Conn, error) {
 			return dbus.Dial("unix:path=/run/systemd/private")
 		})
 	})
@@ -201,7 +201,7 @@ func (c *Conn) GetManagerProperty(prop string) (string, error) {
 	return variant.String(), nil
 }
 
-func dbusAuthConnection(createBus func() (*dbus.Conn, error)) (*dbus.Conn, error) {
+func dbusAuthConnection(createBus func(opts ...dbus.ConnOption) (*dbus.Conn, error)) (*dbus.Conn, error) {
 	conn, err := createBus()
 	if err != nil {
 		return nil, err
@@ -221,7 +221,7 @@ func dbusAuthConnection(createBus func() (*dbus.Conn, error)) (*dbus.Conn, error
 	return conn, nil
 }
 
-func dbusAuthHelloConnection(createBus func() (*dbus.Conn, error)) (*dbus.Conn, error) {
+func dbusAuthHelloConnection(createBus func(opts ...dbus.ConnOption) (*dbus.Conn, error)) (*dbus.Conn, error) {
 	conn, err := dbusAuthConnection(createBus)
 	if err != nil {
 		return nil, err


purpleidea:
Here's the fix for https://github.com/coreos/go-systemd/issues/295

lucab:
This is the same as https://github.com/coreos/go-systemd/pull/293 and https://github.com/coreos/go-systemd/pull/294.

purpleidea:
@lucab Yeah just saw. Well, everything seems to work with git master. Libraries like this shouldn't depend on old version or vendor. Make the updated patch will pass??

lucab:
Indeed, we don't vendor nor depend on old versions here, we keep the peace of whatever is the latest go-dbus release. We test with that and we communicate it to consumers via dep manifest. Usual dependency pinning+vendoring should be done by top-level consumers. @purpleidea where did you get a breakage?

Medium term, I'll be happy to update here and release as soon a new go-dbus release is out. I don't know what @jsouthworth plans are yet though.

purpleidea:
@lucab It broke the builds in https://github.com/purpleidea/mgmt/ but a much easier way to show the breakage is by running: `go get github.com/coreos/go-systemd/` ...

> Usual dependency pinning+vendoring should be done by top-level consumers.

For this type of library, there shouldn't be *any* vendoring, it should work with git master... The broken `go get` experience is the best proof of that.

Why aren't we merging this patch now?? Maybe I'm missing something?

jsouthworth:
Given that this was unintentional, I'll tag a new dbus release now. It didn't occur to me that folks would be passing the functions around so there would be type conflicts, it should have, sorry about that. In other cases (direct calls) this would have been an invisible change since the default case is equivalent.

jsouthworth:
Tagged as v5.0 https://github.com/godbus/dbus/tree/v5.0

purpleidea:
Okay, fixed patch updated with new version. Although I don't think we should have the toml file.

purpleidea:
I've added a second patch to modernize travis. If you'd like them squished, I can do that, but I think it's better this way. I pushed in reverse order so they both should pass.

purpleidea:
(As an aside, I think the GH webui has a bug, because in my tree the travis patch comes first, although they are displaying backwards. Neat bug!)

purpleidea:
@lucab \o/ Tests passed, ready for merge =D

.travis.yml
```diff
@@ -5,8 +5,8 @@ services:
 
 language: go
 go:
-  - "1.5.x"
   - "1.10.x"
+  - "1.11.x"
```
lucab: The lower of this two entries is to check the minimum toolchain we support (mostly for distro packaging and LTS products). I think after this PR the new minimum is 1.7.x. Can you please update both this and the README in that sense?

purpleidea: Sure, all updated and pushed. I kept 1.11 in because it makes sense to test the latest stable as well.

lucab:
@jsouthworth thanks!

@purpleidea thanks for the fixes. The two commits are fine, I just have a minor nit, after which we can merge. FWIW `mgmt` vendor directory is missing both `coreos/go-systemd` and `godbus/dbus`, which is the reason why it broke here and why it is likely to suddenly break again in the future.

.travis.yml
lucab: Yup. You can drop 1.10 too if you feel like it.

purpleidea:
> @purpleidea thanks for the fixes.

np

> The two commits are fine, I just have a minor nit

What is it?

> FWIW mgmt vendor directory is missing both coreos/go-systemd and godbus/dbus, which is the reason why it broke here and why it is likely to suddenly break again in the future.

I respectfully disagree. I believe that responsible libraries should always keep git master stable, and if there really is a reason to break API, then to ideally warn projects, and do it quickly. In any case, if I had vendored my deps, I wouldn't know what broke until way later down the road, and then finding and fixing the issue would be more expensive.

In the history of mgmt, we've only had builds break because of git master changes about four times. And all we're fixed right away, either by a small fix on our side, in the upstream, and in at least one case we caught a bug that a dep didn't catch in their CI.

I probably would have caught this break earlier except I've been AFK a bit lately.
The only dep we vendor is the weirdo deps for etcd grpc since those projects aren't responsible about git master and I can't replace that :(

Hopefully I can convince you to move to pure git master =D

.travis.yml
purpleidea: I'd like to keep N-1 please. It's what I and many people who aren't on the latest distro still use. I drop the versions when there's no stable Fedora version using them anymore. (My personal rule of thumb anyways.)

lucab:
Ack. The nit was on the Travis thing. I understand your point and the whole thing is a bit blurry in Go-land. LGTM, merging.

purpleidea:
\o/

peterfraedrich:
I'm still getting `dep` errors when importing `coreos/go-systemd` because there has been no new tag with this `godbus/dbus` fix yet. Can someone please cut a new tag (v18 i guess?)?
```
Fetching sources...

Solving failure: No versions of github.com/coreos/go-systemd met constraints:
	v17: Could not introduce github.com/coreos/go-systemd@v17, as its subpackage github.com/coreos/go-systemd does not contain usable Go code (*build.NoGoError).. (Package is required by (root).)
	v16: Could not introduce github.com/coreos/go-systemd@v16, as its subpackage github.com/coreos/go-systemd does not contain usable Go code (*build.NoGoError).. (Package is required by (root).)
	v15: Could not introduce github.com/coreos/go-systemd@v15, as its subpackage github.com/coreos/go-systemd does not contain usable Go code (*build.NoGoError).. (Package is required by (root).)
	v14: Could not introduce github.com/coreos/go-systemd@v14, as its subpackage github.com/coreos/go-systemd does not contain usable Go code (*build.NoGoError).. (Package is required by (root).)
	v13: Could not introduce github.com/coreos/go-systemd@v13, as its subpackage github.com/coreos/go-systemd does not contain usable Go code (*build.NoGoError).. (Package is required by (root).)
	v12: Could not introduce github.com/coreos/go-systemd@v12, as its subpackage github.com/coreos/go-systemd does not contain usable Go code (*build.NoGoError).. (Package is required by (root).)
	v11: Could not introduce github.com/coreos/go-systemd@v11, as its subpackage github.com/coreos/go-systemd does not contain usable Go code (*build.NoGoError).. (Package is required by (root).)
	v10: Could not introduce github.com/coreos/go-systemd@v10, as its subpackage github.com/coreos/go-systemd does not contain usable Go code (*build.NoGoError).. (Package is required by (root).)
	v9: Could not introduce github.com/coreos/go-systemd@v9, as its subpackage github.com/coreos/go-systemd does not contain usable Go code (*build.NoGoError).. (Package is required by (root).)
	v8: Could not introduce github.com/coreos/go-systemd@v8, as its subpackage github.com/coreos/go-systemd does not contain usable Go code (*build.NoGoError).. (Package is required by (root).)
	v7: Could not introduce github.com/coreos/go-systemd@v7, as its subpackage github.com/coreos/go-systemd does not contain usable Go code (*build.NoGoError).. (Package is required by (root).)
	v6: Could not introduce github.com/coreos/go-systemd@v6, as its subpackage github.com/coreos/go-systemd does not contain usable Go code (*build.NoGoError).. (Package is required by (root).)
	v5: Could not introduce github.com/coreos/go-systemd@v5, as its subpackage github.com/coreos/go-systemd does not contain usable Go code (*build.NoGoError).. (Package is required by (root).)
	v4: Could not introduce github.com/coreos/go-systemd@v4, as its subpackage github.com/coreos/go-systemd does not contain usable Go code (*build.NoGoError).. (Package is required by (root).)
	v3: Could not introduce github.com/coreos/go-systemd@v3, as its subpackage github.com/coreos/go-systemd does not contain usable Go code (*build.NoGoError).. (Package is required by (root).)
	v2: Could not introduce github.com/coreos/go-systemd@v2, as its subpackage github.com/coreos/go-systemd does not contain usable Go code (*build.NoGoError).. (Package is required by (root).)
	v1: Could not introduce github.com/coreos/go-systemd@v1, as its subpackage github.com/coreos/go-systemd does not contain usable Go code (*build.NoGoError).. (Package is required by (root).)
	0: Could not introduce github.com/coreos/go-systemd@0, as its subpackage github.com/coreos/go-systemd does not contain usable Go code (*build.NoGoError).. (Package is required by (root).)
	master: Could not introduce github.com/coreos/go-systemd@master, as its subpackage github.com/coreos/go-systemd does not contain usable Go code (*build.NoGoError).. (Package is required by (root).)
```

peterfraedrich:
opened #297 to track new release tag

purpleidea:
Stop using dep and just use master. It works great.
If not use a git submodule in vendor/ and point it to the exact tag
you want. (Currently git master for both.)

purpleidea:
On Tue, Oct 30, 2018 at 5:22 PM Peter Christian Fraedrich
<notifications@github.com> wrote:
>
> @purpleidea without explicit deprecation of dep support the fix shouldn't be don't use dep, it should to be cut a new release tag. if the repo maintainers want to deprecate their support for dep then it should be communicated in advance and at the very least with the last dep release being stable

I understand, and probably even agree. Just making a recommendation
that hopefully helps. Good luck!

h8liu:
btw, same issue for github.com/coreos/container-linux-config-transpiler, which vendors `go-systemd` packages but not `dbus` packages.

lucab:
@h8liu as far as I can tell, `container-linux-config-transpiler` should have no problems as it doesn't consume `go-systemd/dbus`. Travis seems to agree as the current master is green. If you are seeing problems building that, can you please report it at https://github.com/coreos/bugs? It may be something being off in your GOPATH, or a breakage in newer toolchain.

h8liu:
@lucab , fyi, here is the build that is failing on gopkgs.io:
https://gopkgs.io/builds/20181102-060850-8e2dbf73

gopkgs.io tries to build every package in a git repo. the package is not actually imported by anything from the repo though, but it is vendored. it might be an bug/issue of the vendor manager that the repo is using..

PR #298, opened 2018-12-13
dbus: Add GetUnitByPID

APTy:
This implements the `org.freedesktop.systemd1.Manager.GetUnitByPID` method from https://www.freedesktop.org/wiki/Software/systemd/dbus/

The test is a very rough outline - I couldn't get it to run locally. Any tips on that?

### DIFF
--- a/dbus/methods.go
+++ b/dbus/methods.go
@@ -288,6 +288,15 @@ func (c *Conn) listUnitsInternal(f storeFunc) ([]UnitStatus, error) {
 	return status, nil
 }
 
+// GetUnitByPID returns an array with all currently loaded units. Note that
+func (c *Conn) GetUnitByPID(pid int) (dbus.ObjectPath, error) {
+	var op dbus.ObjectPath
+	if err := c.sysobj.Call("org.freedesktop.systemd1.Manager.GetUnitByPID", 0, uint(pid)).Store(&op); err != nil {
+		return dbus.ObjectPath(""), err
+	}
+	return op, nil
+}
+
 // ListUnits returns an array with all currently loaded units. Note that
 // units may be known by multiple names at the same time, and hence there might
 // be more unit names loaded than actual units behind them.
--- a/dbus/methods_test.go
+++ b/dbus/methods_test.go
@@ -16,11 +16,14 @@ package dbus
 
 import (
 	"fmt"
+	"io/ioutil"
+	"log"
 	"os"
 	"os/exec"
 	"path"
 	"path/filepath"
 	"reflect"
+	"strconv"
 	"syscall"
 	"testing"
 	"time"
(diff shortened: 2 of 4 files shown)


dbus/methods.go
```diff
@@ -288,6 +288,15 @@ func (c *Conn) listUnitsInternal(f storeFunc) ([]UnitStatus, error) {
 	return status, nil
 }
 
+// GetUnitByPID returns an array with all currently loaded units. Note that
```
lucab: There is part of a sentence missing after "Note that".

dbus/methods.go
```diff
@@ -288,6 +288,15 @@ func (c *Conn) listUnitsInternal(f storeFunc) ([]UnitStatus, error) {
...
+func (c *Conn) GetUnitByPID(pid int) (dbus.ObjectPath, error) {
```
lucab: Input parameter should be a `uint32` to match DBus signature.

dbus/methods_test.go
```diff
@@ -132,6 +135,18 @@ func runStopUnit(t *testing.T, conn *Conn, trTarget TrUnitProp) error {
 	return nil
 }
 
+func assertNoError(t *testing.T, err error) {
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+func assertEqualStr(t *testing.T, shouldBe, target string) {
```
lucab: It is ok to inline this and the helper above directly inside the test function.

fixtures/get-unit-pid.sh
```diff
@@ -0,0 +1,2 @@
+/bin/echo ${PID} >get-unit-pid.pid
```
lucab: This should have a `#!/usr/bin/env bash` shebang.

fixtures/get-unit-pid.service
```diff
@@ -0,0 +1,5 @@
+[Unit]
+Description=get unit pid
+
+[Service]
+ExecStart=get-unit-pid.sh
```
lucab: This should specify the full absolute path to the binary.

lucab:
Thanks for the PR! Code looks mostly ok.
I think that test is failing because you didn't specify an absolute path for the binary.

If path lookup gets too complex, you can try re-arranging the test so that it just runs a sleeping service.  That way you can get the service by name, then lookup its main PID from properties, then use your method with the PID to get the path, and finally verify that it matches with the service property.

dbus/methods.go
APTy: ack

dbus/methods.go
APTy: I agree. I was thinking of using `int` because its what the go stdlib uses for pids (oddly enough), but I'd prefer to adhere to the dbus signature

APTy:
> That way you can get the service by name, then lookup its main PID from properties, then use your method with the PID to get the path, and finally verify that it matches with the service property.

I'm having trouble looking up the pid from properties with `GetUnitProperties` or `GetUnitProperty`. Is it "MainPID" that I should be looking for?

APTy:
Got it, thanks. Do you happen to know if it's only available after certain versions of systemd? I tried using `GetUnitProperty` to look it up (and could find other properties like `Id`) but couldn't find `MainPID`. I can follow up with some repro steps but just wanted to check first

APTy:
Seems like the call to `GetUnitByPID()` on 18.04 is returning the object path to the root slice configuration unit `-.slice` (http://manpages.ubuntu.com/manpages/bionic/man5/systemd.slice.5.html), looking into this more

kolyshkin:
Closing in favor of #385 (which was merged in 2021).

Issue #290, opened 2018-10-04
flake: sporadic test failures from TestMachine


lucab:
`machine1` testsuite sporadically fails with:

```
=== RUN   TestMachine
--- FAIL: TestMachine (0.03s)
	dbus_test.go:102: unexpectedly found machine named machined-test-mykjbyxx
```

I think this could be because `TerminateMachine` is asynchronous and we are racing against termination, but I didn't check server-side code in systemd to confirm that. If that is the case, testcode should be augmented with retry+timeout logic.

PR #299, opened 2019-02-01
journal: skip sending test if journald is unavailable

pkern:
TestJournalSend tries to communicate with the local journald and send some messages there. There are environments where go-systemd is useful but there is no local journald available. In these cases it should still be possible to run the test suite successfully.


pkern:
The one failing test unfortunately looks bogus to me. I don't think my change here caused that one.

lucab:
Thanks for the PR! That was indeed the flake at https://github.com/coreos/go-systemd/issues/290. CI is green after re-triggering it.

PR #301, opened 2019-02-11
test/machine1: fix sporadic test failures from TestMachine

### LINKED ISSUE #290: flake: sporadic test failures from TestMachine

LorbusChris:
Added timeout/retry logic to async `TerminateMachine` test

Resolves #290

### DIFF
--- a/machine1/dbus_test.go
+++ b/machine1/dbus_test.go
@@ -97,7 +97,13 @@ func TestMachine(t *testing.T) {
 		t.Fatal(tErr)
 	}
 
-	machine, getErr = conn.GetMachine(machineName)
+	for i := 1; i <= 10; i++ {
+		machine, getErr = conn.GetMachine(machineName)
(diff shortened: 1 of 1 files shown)


LorbusChris:
CI failed with Go 1.11 on Ubuntu 16.04. Another flake?

cc @lucab

lucab:
Likely so, on `TestStartStopTransientUnitAll`. But it is the first time I'm seeing that, I'm not very concerned right now.

PR #303, opened 2019-03-17
examples: Listeners method is now without arguments

gudata:


lucab:
Indeed, thanks for the PR!

PR #304, opened 2019-03-17
examples: systemd-activate is now systemd-socket-activate

gudata:
systemd-activate is now systemd-socket-activate


lucab:
Good catch, LGTM!

PR #306, opened 2019-05-03
dbus: Introduce GetAllProperties() for fetching all systemd unit properties

martynasb:
Hi,

I would like to fetch a bunch of systemd service unit properties (e.g. NRestarts, ActiveState, SubState, etc.) in a single call, similar to what I get by running `systemctl show <service>`.

Currently, `dbus/methods.go` exposes two methods for retrieving unit properties:

```
GetUnitTypeProperties(unit string, unitType string)
|-> return c.getProperties(path, "org.freedesktop.systemd1."+unitType) // unitType = Service

GetUnitProperties(unit string)
|-> return c.getProperties(path, "org.freedesktop.systemd1.Unit")
```

Both of these methods call the `getProperties()` helper method:

```
getProperties(path dbus.ObjectPath, dbusInterface string)
|   ...
|-> err = obj.Call("org.freedesktop.DBus.Properties.GetAll", 0, dbusInterface).Store(&props)
```

`getProperties()` will return a different set of properties depending on the `dbusInterface` parameter value (i.e. `org.freedesktop.systemd1.Unit` vs `org.freedesktop.systemd1.Service`). See the `dbus-send` examples below that illustrate this:

```
$ dbus-send --system --dest=org.freedesktop.systemd1 --type=method_call /org/freedesktop/systemd1/unit/<MYSERVICE>_2eservice --print-reply org.freedesktop.DBus.Properties.GetAll string:"org.freedesktop.systemd1.Service" | grep -A 1 -e NRestarts -e ActiveState -e SubState
         string "NRestarts"
         variant             uint32 0
```

```
$ dbus-send --system --dest=org.freedesktop.systemd1 --type=method_call /org/freedesktop/systemd1/unit/<MYSERVICE>_2eservice --print-reply org.freedesktop.DBus.Properties.GetAll string:"org.freedesktop.systemd1.Unit" | grep -A 1 -e NRestarts -e ActiveState -e SubState
         string "ActiveState"
         variant             string "active"
--
         string "SubState"
         variant             string "running"
```

However, if we set the interface argument to an empty string, we get all of these properties in a single call:

```
$ dbus-send --system --dest=org.freedesktop.systemd1 --type=method_call /org/freedesktop/systemd1/unit/<MYSERVICE>_2eservice --print-reply org.freedesktop.DBus.Properties.GetAll string:"" | grep -A 1 -e NRestarts -e ActiveState -e SubState
         string "NRestarts"
         variant             uint32 0
--
         string "ActiveState"
         variant             string "inactive"
--
         string "SubState"
         variant             string "dead"
```

I would like to be able to do the same using this lib by invoking something along the lines of `dbus.GetProperties("<MYSERVICE>.service", "")`

### DIFF
--- a/dbus/methods.go
+++ b/dbus/methods.go
@@ -197,6 +197,12 @@ func (c *Conn) GetUnitPathProperties(path dbus.ObjectPath) (map[string]interface
 	return c.getProperties(path, "org.freedesktop.systemd1.Unit")
 }
 
+// GetAllProperties takes the (unescaped) unit name and returns all of its dbus object properties.
+func (c *Conn) GetAllProperties(unit string) (map[string]interface{}, error) {
+	path := unitPath(unit)
+	return c.getProperties(path, "")
+}
+
 func (c *Conn) getProperty(unit string, dbusInterface string, propertyName string) (*Property, error) {
 	var err error
 	var prop dbus.Variant


lucab:
Thanks for the PR! I'm currently away from my usual workstation; I'll give a look at this once I'm back 
(with the rest of the backlog), likely in a few days.

martynasb:
ping

lucab:
Sorry for the long delay. I'm fine with adding this feature to the library.

The only comment I have is that if the actual goal of the PR is to expose all-interfaces properties, I'd directly call this `GetAllProperties` and get rid of the last string parameter. @martynasb do you have a usecase where you need to pass a non-empty arbitrary string, which is not already covered by the other methods?

martynasb:
> The only comment I have is that if the actual goal of the PR is to expose all-interfaces properties, I'd directly call this `GetAllProperties` and get rid of the last string parameter.

I've updated my PR accordingly.

> @martynasb do you have a usecase where you need to pass a non-empty arbitrary string, which is not already covered by the other methods?

Nope. `GetAllProperties()` should cover our all use cases.

lucab:
Great, thanks! Please just squash the commits, and I'll approve and merge.

martynasb:
Thanks. Can you cut a new release?

lucab:
Yep, but I'd like to get https://github.com/coreos/go-systemd/pull/310 in first. It shouldn't take too long, I'll ping back here once done.

lucab:
https://github.com/coreos/go-systemd/releases/tag/v20

martynasb:
Thanks!

PR #307, opened 2019-05-15
Add go module support

saschagrunert:
This PR adds a `go.mod` and `go.sum` to the project which contains the latest dependencies. It would be awesome if we could tag a release according to semver after this PR got merged, WDYT? :innocent:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,14 +5,15 @@ services:
 
 language: go
 go:
-  - "1.10.x"
-  - "1.11.x"
+  - "1.12.x"
+  - "1.13.x"
 go_import_path: github.com/coreos/go-systemd
 
 env:
   global:
     - IMPORTPATH=github.com/coreos/go-systemd
     - GOPATH=/opt
+    - GO111MODULE=on
     - DEP_BINDIR=/tmp
     - BUILD_DIR=/opt/src/github.com/coreos/go-systemd
   matrix:
@@ -23,8 +24,6 @@ env:
 before_install:
  - sudo apt-get -qq update
  - sudo apt-get install -y libsystemd-journal-dev libsystemd-daemon-dev
- - curl https://raw.githubusercontent.com/golang/dep/master/install.sh | INSTALL_DIRECTORY=${DEP_BINDIR} sh
- - ${DEP_BINDIR}/dep ensure -v
  - docker pull ${DOCKER_BASE}
  - docker run --privileged -e GOPATH=${GOPATH} --cidfile=/tmp/cidfile ${DOCKER_BASE} /bin/bash -c "apt-get update && apt-get install -y sudo build-essential git golang dbus libsystemd-dev libpam-systemd systemd-container"
  - docker commit `cat /tmp/cidfile` go-systemd/container-tests
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,5 @@
+module github.com/coreos/go-systemd/v21
+
+go 1.12
+
+require github.com/godbus/dbus/v5 v5.0.3
--- a/go.sum
+++ b/go.sum
@@ -1,0 +1,2 @@
+github.com/godbus/dbus/v5 v5.0.3 h1:ZqHaoEF7TBzh4jzPmqVhE/5A1z9of6orkAe5uHoAeME=
+github.com/godbus/dbus/v5 v5.0.3/go.mod h1:xhWf0FNVPg57R7Z0UbKHbJfkEywrmjJnf7w5xrFpKfA=
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+test_bins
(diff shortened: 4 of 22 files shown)


lucab:
We touched this topic in the past with @arithx, and we are a bit cautious.

Relevant sharp edges:
 * there is no overall `go-systemd` package, just individual sub-packages. Versioning is shared, though.
 * this repository had tagged releases since the beginning, but gomod doesn't understand the format. So the starting version likely has to be with `v20.0.0` (next tag).
 * the API re-exports types from other libraries (e.g. godbus). As such, I expect major-version bumps to be the norm (as it is currently done).
 * related to the above, I'd like not to have to move directories around for each breaking change (i.e. `/vX` -> `/vY`). I think this can/should be done via the top-level `module` definition.
 * gomod support in libraries consumed here still needs some adjustments, see https://github.com/godbus/dbus/issues/125

Xuanwo:
Maybe we can switch to go mod in next release? 

Use `https://github.com/coreos/go-systemd/v20` as module name just like `https://github.com/google/go-github`.

lucab:
I'd like to do a couple of first-hand experiments before that, in particular wrt. subpackages and module-version, so likely not targeting this at the next release.
And yes, puttting the version label in the `module` directive was my approach in the above list.

lucab:
Also relevant here: go-systemd consumes `coreos/pkg` which is a mix bag of messy thing and not very curated. But, we strictly only need `coreos/pkg/dlopen`, which is quite stable and easier to version.

I'll check and arrange to get the latter moved to `coreos/dlopen` next, with its own semver and go module.

lucab:
Additionally, it is my current understanding that we have to rewrite all our internal imports to be in the form `github.com/coreos/go-systemd/vX/sub_package` and keep the `vX` portion in sync with the `module` declaration.

Xuanwo:
Yes, we need to update all files while upgrading major version.

Golang may add `go mod upgrade` to help reduce the work, but not accepted for now.

saschagrunert:
Hey, any idea how to move forward here? The latest version of godbus/dbus is now using a v5 import suffix which makes it hard to use go-systemd with projects using go modules.

lucab:
@saschagrunert thanks for pinging back. I think we can more or less stick to the plan highlighted so far.
In particular, I think we can proceed as follows:
 * copy `dlopen` under an `internal/` hierarchy here and adjust imports
 * release `v21` with that, which would be the last version using `dep`
 * swap manifests dep->gomod, and introduce `v21` in all imports
 * bump everything to `v22` and release a `v22.0.0` tag

I don't have any of this on my current radar as I'm unfortunately already fully-booked on other topics, but I'll gladly review any PR and try to quickly cut releases when ready. If you have a pressing timeline on this, feel free to start PRing.

ulidtko:
What a coincidence; I'm currently struggling with exactly this issue, too.

I don't feel proficient enough in Go to roll a PR, so the only thing left for me is to wait. Sigh...

go.sum
```diff
@@ -0,0 +1,4 @@
+github.com/godbus/dbus v4.1.0+incompatible h1:WqqLRTsQic3apZUK9qC5sGNfXthmPXzUZ7nQPrNITa4=
+github.com/godbus/dbus v4.1.0+incompatible/go.mod h1:/YcGZj5zSblfDWMMoOzV4fas9FZnQYTkDnsGvmh2Grw=
```
lucab: This line and the one above look spurious, no?

saschagrunert: Yes, I will remove it.

saschagrunert:
Would it be fine if we drop the build for go 1.10? Because the v5 imports won't work with that.

saschagrunert:
PTAL

.travis.yml
```diff
@@ -5,14 +5,15 @@ services:
 
 language: go
 go:
-  - "1.10.x"
```
lucab: This is fine, but it the badge in the README needs to be updated as well.

go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/coreos/go-systemd
```
lucab: I think we need to introduce the `/v21` here, no?

.travis.yml
saschagrunert: Yep, I'll adapt.

go.mod
saschagrunert: Ah yes, I'll change that and adapt the imports.

chrboe:
Is there some kind of workaround for people trying to use this package in a module environment, or will we just have to wait around for this to get merged?

saschagrunert:
> Is there some kind of workaround for people trying to use this package in a module environment, or will we just have to wait around for this to get merged?

I right now stick to an older version.

@lucab do you think we can move forward here?

lucab:
Ah, I didn't realize this was re-pushed after my last review. LGTM now, I'll merge.

chrboe:
Wow, that was quick! Thanks

lucab:
Tagged: https://github.com/coreos/go-systemd/releases/tag/v22.0.0

PR #308, opened 2019-06-12
sdjournal: Provide error messages as strings

stevenjm:
This was previously just displaying the numeric errno value, resulting in error messages like:

```failed to seek to cursor "foo\n": 22```

The `syscall.Errno` type provides a handy function for returning the string value of an error, so just use that for more informative messages.


lucab:
Thanks for the PR! Just for clarity, this wouldn't provide any more information about the error, right? It only does the equivalent of `strerror(3)` (errno int->string). Was that your goal?

stevenjm:
Correct, and indeed it is my goal. By "more informative" I meant that the message is more informative at a glance, not that there's additional information supplied.

lucab:
@stevenjm ack, LGTM!

Issue #232, opened 2017-03-26
Systemd-Importd support


cpg1111:
Hi there,

Would there be any opposition to implementing support for importd? [importd docs found here](https://www.freedesktop.org/wiki/Software/systemd/importd/).

I'd be happy to implement this and submit a pull request is this is something desired.

Thanks

lucab:
I haven't personally heard of many requests for importd support, but if you have usecases for it we'll gladly review and merge here.

As this systemd-importd is usually not enabled on most distros, please add some runtime detection for tests (so they can be loudly-skipped where support is missing).

Xuanwo:
I'm interested on this too, I will implement it in later few days.

PR #310, opened 2019-06-17
import1: Implement importd methods

Xuanwo:
systemd-importd service is used to import, export and download
additional system images, for running them as local containers
using tools such as systemd-nspawn.

This PR allows users to import/export/pull images via
systemd-importd.

Fix https://github.com/coreos/go-systemd/issues/232


Xuanwo:
ping @lucab, this PR is ready for review.

import1/dbus.go
```diff
@@ -0,0 +1,220 @@
...
+	c := new(Conn)
+
+	if err := c.initConnection(); err != nil {
+		return nil, err
+	}
+
+	return c, nil
+}
+
+func (c *Conn) initConnection() error {
```
lucab: Note to self: this is a connection helper duplicated in a few places, we should later move it to a common sub-package under `internal`.

import1/dbus.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+// Transfer is the status for an import, export or download operations.
+type Transfer struct {
+	Id       uint32  // The numeric transfer ID of the transfer object
+	Local    string  // The local container name of this transfer
+	Remote   string  // The remote source (in case of download: the URL, in case of import/export a string describing the file descriptor passed in)
+	Type     string  // The type of operation
+	Verify   string  // The selected verification setting, and is only defined for download operations
+	Progress float64 // The current progress of the transfer, as a value between 0.0 and 1.0
```
lucab: Is there a reason to have these as side-comments instead of normal docstrings?

import1/dbus.go
```diff
@@ -0,0 +1,220 @@
...
+		return err
+	}
+
+	c.object = c.conn.Object("org.freedesktop.import1", dbus.ObjectPath(dbusPath))
+
+	return nil
+}
+
+func (c *Conn) getResult(method string, args ...interface{}) (
+	transfer_id uint32, transfer_path dbus.ObjectPath, err error,
```
lucab: I'd prefer to avoid using named-arguments as return values. Explicit return is a bit easier to read and maintain.

import1/dbus.go
```diff
@@ -0,0 +1,220 @@
...
+) {
+	result := c.object.Call(fmt.Sprintf("%s.%s", dbusInterface, method), 0, args...)
+	if result.Err != nil {
+		return 0, "", result.Err
+	}
+
+	transfer_id, ok := result.Body[0].(uint32)
```
lucab: Please check that `len(resul.Body) >= 2` first.

import1/dbus.go
```diff
@@ -0,0 +1,220 @@
...
+	return transfers, nil
+}
+
+// CancelTransfer will cancel an ongoing import, export or download operations.
+func (c *Conn) CancelTransfer(transfer_id uint32) error {
+	return c.object.Call(dbusInterface+".CancelTransfer", 0, transfer_id).Err
+}
+
+func transferFromInterfaces(transfer []interface{}) (*Transfer, error) {
+	// Verify could be not defined in response.
```
lucab: s/could/may/

import1/dbus.go
```diff
@@ -0,0 +1,220 @@
...
+	if !ok {
+		return 0, "", fmt.Errorf("unable to convert dbus response '%v' to dbus.ObjectPath", result.Body[1])
+	}
+	return
+}
+
+// ImportTar imports a tar into systemd-importd.
+func (c *Conn) ImportTar(
+	fd dbus.UnixFD, local_name string, force, read_only bool,
+) (uint32, dbus.ObjectPath, error) {
```
lucab: Here an everywhere below, it looks like this `(uint32, dbus.ObjectPath)` would be better returned as a single `TransferObejct`.

import1/dbus.go
```diff
@@ -0,0 +1,220 @@
...
+	transfer_path, ok = result.Body[1].(dbus.ObjectPath)
+	if !ok {
+		return 0, "", fmt.Errorf("unable to convert dbus response '%v' to dbus.ObjectPath", result.Body[1])
+	}
+	return
+}
+
+// ImportTar imports a tar into systemd-importd.
+func (c *Conn) ImportTar(
+	fd dbus.UnixFD, local_name string, force, read_only bool,
```
lucab: Here and below, what do you think of taking an `*os.File` here and doing the fd-conversion internally?

import1/dbus.go
Xuanwo: Well, just learnt from `machine1`. I have no idea on this and will update them if requested.

import1/dbus.go
Xuanwo: OK, I will update it.

import1/dbus.go
Xuanwo: Firstly, I'm trying to keep fit with importd's dbus API just like other APIs.
Secondly, there are only `Id` and `dbus.ObjectPath` here, do you mean that we need to query the complete `TransferObejct` from `sd-imported`?

import1/dbus.go
Xuanwo: Yep, I do agree with this. I will refactor those APIs.

import1/dbus.go
lucab: Ack, I'm fine with keeping those as they are now.

import1/dbus.go
Xuanwo: Got it.

import1/dbus_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+	_, err = os.Stat(filepath)
+	if err == nil {
+		return
+	}
+	if !os.IsNotExist(err) {
+		t.Fatal(err)
+	}
+
+	res, err := http.Get(url)
```
lucab: Do we need a full-image and a remote download here? Can we maybe just have a small `.raw.xz` and `.tar.gz` under `/fixtures/`?

import1/dbus.go
```diff
@@ -0,0 +1,235 @@
...
+		c.conn.Close()
+		return err
+	}
+
+	c.object = c.conn.Object("org.freedesktop.import1", dbus.ObjectPath(dbusPath))
+
+	return nil
+}
+
+// getResult will return:
```
lucab: This docstring needs to be updated for `*Transfer`.

lucab:
Two minor things left, looks ready otherwise. @Xuanwo feel free to squash into a single commit.

import1/dbus_test.go
Xuanwo: Well, I used to think systemd will check validity of the image, but I found I was wrong. I will add fixtures instead.

import1/dbus.go
Xuanwo: Yep, I forgot to update here.

Xuanwo:
This PR is ready for merge now :smile:

PR #312, opened 2019-07-09
machine1: Implement CloneImage

Xuanwo:
This commit added CloneImage support in dbus, user can clone their
images via dbus API.

### DIFF
--- a/machine1/dbus_test.go
+++ b/machine1/dbus_test.go
@@ -157,7 +157,10 @@ func generateRandomLabel(n int) string {
 
 func TestImages(t *testing.T) {
 	imageName := machinePrefix + generateRandomLabel(8)
-	imagePath := filepath.Join("/var/lib/machines", imageName)
+	imagePath := filepath.Join("/var/lib/machines", imageName+".raw")
+
+	newImageName := machinePrefix + generateRandomLabel(8)
+	newImagePath := filepath.Join("/var/lib/machines", newImageName+".raw")
 
 	if _, err := os.Create(imagePath); err != nil {
 		t.Fatal(err)
@@ -180,5 +183,15 @@ func TestImages(t *testing.T) {
 
 	if len(listImages) < 1 {
 		t.Fatalf("did not find any image")
+	}
+
+	cloneErr := conn.CloneImage(imageName, newImageName, false)
+	if cloneErr != nil {
+		t.Fatal(cloneErr)
+	}
+	defer os.Remove(newImagePath)
+
+	if len(listImages) < 2 {
+		t.Fatalf("image clone failed")
 	}
 }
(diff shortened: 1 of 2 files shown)


machine1/dbus_test.go
```diff
@@ -157,7 +157,10 @@ func generateRandomLabel(n int) string {
 
 func TestImages(t *testing.T) {
 	imageName := machinePrefix + generateRandomLabel(8)
-	imagePath := filepath.Join("/var/lib/machines", imageName)
+	imagePath := filepath.Join("/var/lib/machines", imageName+".raw")
+
+	newImageName := machinePrefix + generateRandomLabel(8)
+	newImagePath := filepath.Join("/var/lib/machines", newImageName+".raw")
```
lucab: Is this used anywhere?

machine1/dbus_test.go
```diff
@@ -181,4 +184,14 @@ func TestImages(t *testing.T) {
 	if len(listImages) < 1 {
 		t.Fatalf("did not find any image")
 	}
+
+	cloneErr := conn.CloneImage(imageName, newImageName, false)
+	if cloneErr != nil {
+		t.Fatal(cloneErr)
+	}
+	defer os.Remove(newImagePath)
```
lucab: I think you may want to `RemoveImage` here, not removing by FS path.

lucab:
@Xuanwo thanks for the PR! I've left a few comments regarding the testing logic.

machine1/dbus_test.go
Xuanwo: It's used in CloneImage test.

machine1/dbus_test.go
Xuanwo: The logic is borrowed from old tests, maybe we should refactor them all?

machine1/dbus_test.go
lucab: I'm sorry, I'm not following. `newImagePath` is used exactly once in a `os.Remove`; who is creating and populating that file?

Xuanwo: conn.CloneImage will copy the image to new place.

machine1/dbus_test.go
lucab: The old test is creating the file directly, while here it happens implicitly (by machined). That wasn't clear to me first at first.

If you have time, I'd be happy to see this test doing the following before exiting:
 * doing `defer os.Remove()` on `imagePath` and `newImagePath`
 * calling `RemoveImage` on both `imageName` and `newImageName`
 * checking that  `imagePath` and `newImagePath` do not exist anymore

Xuanwo: OK, I will refactor the logic for test.

PR #313, opened 2019-07-19
activation: fix tautological error handling

muesli:
err can't have changed its value and we would have aborted if it's not nil, above.


muesli:
No problem, thanks for merging!

PR #318, opened 2019-11-01
sdjournal: use internal dlopen package

saschagrunert:
The first step for go modules as discussed here: https://github.com/coreos/go-systemd/pull/307#issuecomment-548800039

cc @lucab 

Afterwards I would need assistance in releasing v21 :innocent:

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -1,8 +1,4 @@
 [[constraint]]
-  name = "github.com/coreos/pkg"
-  version = "4.0.0"
-
-[[constraint]]
   name = "github.com/godbus/dbus"
   version = "5.0"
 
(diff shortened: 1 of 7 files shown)


internal/dlopen/dlopen_example.go
```diff
@@ -0,0 +1,56 @@
+// Copyright 2015 CoreOS, Inc.
```
lucab: We can skip this example file entirely.

saschagrunert: Yes, I removed it.

saschagrunert:
> A minor thing, plus `Gopkg.toml` needs to be updated too. Otherwise looks fine.

Thanks for the review, I'll updated `Gopkg.tomll` too.

lucab:
@saschagrunert  https://github.com/coreos/go-systemd/releases/tag/v21

Issue #326, opened 2019-11-26
login1 command wrappers do not return the call error


coderanger:
The pattern used in other packages here is something like `return c.object.Call(...).Err` but the wrapper methods in `login1` just swallow any errors.

lucab:
Thanks for the report.
I agree it would be better to align all those methods to return proper `error` value. If you are interested in doing it, I'll gladly review and merge a PR doing that.

PR #327, opened 2019-12-07
Add Suspend support to login1

daniloqueiroz:


login1/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/daniloqueiroz/go-systemd/login1
```
lucab: `login1` is not a module on its own. This should be dropped.

login1/dbus.go
```diff
@@ -258,3 +258,8 @@ func (c *Conn) Subscribe(members ...string) chan *dbus.Signal {
 func (c *Conn) PowerOff(askForAuth bool) {
 	c.object.Call(dbusInterface+".PowerOff", 0, askForAuth)
 }
+
+// Suspend asks logind for a suspend optionally asking for auth.
+func (c *Conn) Suspend(askForAuth bool) {
```
lucab: It would be better to return the inner `error` here, see https://github.com/coreos/go-systemd/issues/326.

login1/dbus.go
```diff
@@ -258,3 +258,8 @@ func (c *Conn) Subscribe(members ...string) chan *dbus.Signal {
 func (c *Conn) PowerOff(askForAuth bool) {
 	c.object.Call(dbusInterface+".PowerOff", 0, askForAuth)
 }
+
+// Suspend asks logind for a suspend optionally asking for auth.
```
lucab: Suggestion: `Suspend triggers suspension via logind, optionally asking for authentication.`

lucab:
@daniloqueiroz thanks for the PR! I've left a few comments in the review.

PR #330, opened 2020-03-02
unit: deserialize keeps all semantic data

glawler:
The existing unit.Deserialize() function does not contain all the
semantic data in a unit file. Specifically it does not fully
support multiple duplicate sections. It does correctly parse the sections,
but the `[]*UnitOption` data structure does not tell the caller,
in the case of a unit file with duplicate section names, *which*
section the UnitOptions are from.

This may matter.

For example, a `/etc/systemd/network/*.network` file can have multiple
`[Route]` entries. Just returning an array of `Section, Name, Value`
tuples does not tell the caller which `[Route]` a given `Name, Value`
pair belongs to.

Ex:
```
[Route]
Gateway=10.0.100.1
Destination=10.0.0.1/24

[Route]
Gateway=10.0.100.2
Destination=10.0.2.1/24
```

This unit file will generate the following from `unit.Deserialize()`

```
{Section: "Route", Name: "Gateway", Value: "10.0.100.1"}
{Section: "Route", Name: "Destination", Value: "10.0.0.1/24"}
{Section: "Route", Name: "Gateway", Value: "10.0.100.2"}
{Section: "Route", Name: "Destination", Value: "10.0.2.1/24"}
```

With just this information, it is not possible to know which `Gateway`
goes to which `Destination`. Order maybe, but what if there were not two
Gateways? Which `Destination` would go with the `Gateway`?

This patch introduces a new Deserialize function that returns the
following data structures:

```
// UnitEntry is a single line entry in a Unit file.
type UnitEntry struct {
	Name  string
	Value string
}

// UnitSection is a section in a Unit file. The section name
// and a list of entries in that section.
type UnitSection struct {
	Section string
	Entries []*UnitEntry
}
```

The new DeserializeSection() signature is:

```
func DeserializeSections(f io.Reader) ([]*UnitSection, error)
```

And the output from it given the example unit file above is:

```
{Section: Route{Name: Gateway, Value: 10.0.100.1}{Name: Destination, Value: 10.0.0.1/24}}
{Section: Route{Name: Gateway, Value: 10.0.100.2}{Name: Destination, Value: 10.0.2.1/24}}
```

The `Name`, `Value` pairs are directly associated with a specific
section. There is no ambiguity.

There are unit tests added to `unit/deserialize_test.go` that do basic
tests on `DeserializeSections()`. All exising unit tests in that file
are untouched and pass.

### DIFF
--- a/unit/deserialize.go
+++ b/unit/deserialize.go
@@ -43,37 +43,99 @@ var (
 	ErrLineTooLong = fmt.Errorf("line too long (max %d bytes)", SYSTEMD_LINE_MAX)
 )
 
-// Deserialize parses a systemd unit file into a list of UnitOption objects.
-func Deserialize(f io.Reader) (opts []*UnitOption, err error) {
-	lexer, optchan, errchan := newLexer(f)
-	go lexer.lex()
-
-	for opt := range optchan {
-		opts = append(opts, &(*opt))
-	}
-
-	err = <-errchan
-	return opts, err
+// DeserializeOptions parses a systemd unit file into a list of UnitOptions
+func DeserializeOptions(f io.Reader) (opts []*UnitOption, err error) {
+	_, options, err := deserializeAll(f)
+	return options, err
 }
 
-func newLexer(f io.Reader) (*lexer, <-chan *UnitOption, <-chan error) {
-	optchan := make(chan *UnitOption)
+// DeserializeSections deserializes into a list of UnitSections.
+func DeserializeSections(f io.Reader) ([]*UnitSection, error) {
+	sections, _, err := deserializeAll(f)
+	return sections, err
+}
+
+// Deserialize parses a systemd unit file into a list of UnitOptions.
+// Note: this function is deprecated in favor of DeserializeOptions
+// and will be removed at a future date.
+func Deserialize(f io.Reader) (opts []*UnitOption, err error) {
+	return DeserializeOptions(f)
+}
+
+type lexDataType int
+
+const (
+	SEC lexDataType = iota
+	OPT
+)
+
+// lexChanData - support either datatype in the lex channel.
+// Poor man's union data type.
+type lexData struct {
+	Type    lexDataType
+	Option  *UnitOption
+	Section *UnitSection
+}
+
+// deserializeAll deserializes into UnitSections and UnitOptions.
+func deserializeAll(f io.Reader) ([]*UnitSection, []*UnitOption, error) {
+
+	lexer, lexchan, errchan := newLexer(f)
+
+	go lexer.lex()
+
+	sections := []*UnitSection{}
+	options := []*UnitOption{}
+
+	for ld := range lexchan {
+		switch ld.Type {
+		case OPT:
+			if ld.Option != nil {
+				// add to options
+				opt := ld.Option
+				options = append(options, &(*opt))
+
+				// sanity check. "should not happen" as SEC is first in code flow.
+				if len(sections) == 0 {
+					return nil, nil, fmt.Errorf(
+						"Unit file misparse: option before section")
+				}
+
+				// add to newest section entries.
+				s := len(sections) - 1
+				sections[s].Entries = append(sections[s].Entries,
+					&UnitEntry{Name: opt.Name, Value: opt.Value})
+			}
+		case SEC:
+			if ld.Section != nil {
+				sections = append(sections, ld.Section)
+			}
+		}
+	}
+
+	err := <-errchan
+
+	return sections, options, err
+}
+
+func newLexer(f io.Reader) (*lexer, <-chan *lexData, <-chan error) {
+	lexchan := make(chan *lexData)
 	errchan := make(chan error, 1)
 	buf := bufio.NewReader(f)
 
-	return &lexer{buf, optchan, errchan, ""}, optchan, errchan
+	return &lexer{buf, lexchan, errchan, ""}, lexchan, errchan
 }
 
 type lexer struct {
 	buf     *bufio.Reader
-	optchan chan *UnitOption
+	lexchan chan *lexData
 	errchan chan error
 	section string
 }
 
 func (l *lexer) lex() {
 	defer func() {
-		close(l.optchan)
+		close(l.lexchan)
 		close(l.errchan)
 	}()
 	next := l.lexNextSection
@@ -124,6 +186,12 @@ func (l *lexer) lexSectionSuffixFunc(section string) lexStep {
 		garbage = bytes.TrimSpace(garbage)
 		if len(garbage) > 0 {
 			return nil, fmt.Errorf("found garbage after section name %s: %v", l.section, garbage)
+		}
+
+		l.lexchan <- &lexData{
+			Type:    SEC,
+			Section: &UnitSection{Section: section, Entries: []*UnitEntry{}},
+			Option:  nil,
 		}
 
 		return l.lexNextSectionOrOptionFunc(section), nil
@@ -252,7 +320,11 @@ func (l *lexer) lexOptionValueFunc(section, name string, partial bytes.Buffer) l
 		} else {
 			val = strings.TrimSpace(val)
 		}
-		l.optchan <- &UnitOption{Section: section, Name: name, Value: val}
+		l.lexchan <- &lexData{
+			Type:    OPT,
+			Section: nil,
+			Option:  &UnitOption{Section: section, Name: name, Value: val},
+		}
 
 		return l.lexNextSectionOrOptionFunc(section), nil
 	}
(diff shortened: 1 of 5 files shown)


glawler:
One test is failing in coreos Travis. All tests passing in my fork though. Not sure what the problem is.

glawler:
No supports serializing `UnitSections` to support the `UnitSections`  deserialize. This makes it much more convenient to work in `UnitSections` instead of `UnitOptions`. 

Tests added fot this. All tests passing.

unit/section.go
```diff
@@ -0,0 +1,46 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package unit
+
+import ()
```
lucab: Leftover?

unit/serialize.go
```diff
@@ -58,6 +58,31 @@ func Serialize(opts []*UnitOption) io.Reader {
 	return &buf
 }
 
+func SerializeSections(sections []*UnitSection) io.Reader {
```
lucab: Missing docstring.

unit/serialize.go
```diff
@@ -58,6 +58,31 @@ func Serialize(opts []*UnitOption) io.Reader {
...
+
+	var buf bytes.Buffer
+
+	if len(sections) == 0 {
```
lucab: I think we don't need this, as an empty loop will have the same result.

unit/deserialize.go
```diff
@@ -43,36 +43,83 @@ var (
...
+}
+
+// DeserializeSections deserializes into a list of UnitSections.
+func DeserializeSections(f io.Reader) ([]*UnitSection, error) {
+	sections, _, err := DeserializeAll(f)
+	return sections, err
+}
+
+// DeserializeAll deserializes into UnitSections and UnitOptions.
+func DeserializeAll(f io.Reader) ([]*UnitSection, []*UnitOption, error) {
```
lucab: I am having a bit of an hard time understanding how you are supposed to use this API.
Are the two slice supposed to contain almost the same information, but in a different format (sections vs options)? What's the intended usecase for this?

unit/deserialize.go
```diff
@@ -43,36 +43,83 @@ var (
...
+		select {
+		case opt, ok := <-optchan:
+			if !ok {
+				done = true
+			} else {
+				// add to options
+				options = append(options, &(*opt))
+
+				// add to newest section entries.
+				s := len(sections) - 1
```
lucab: I fear this will panic if we manager to get here with `len == 0`.

unit/deserialize.go
```diff
@@ -43,36 +43,83 @@ var (
...
 	go lexer.lex()
 
-	for opt := range optchan {
-		opts = append(opts, &(*opt))
+	sections := []*UnitSection{}
+	options := []*UnitOption{}
+
+	var done bool
+
+	for {
```
lucab: At an high level, this logic seems problematic.
There is an implicit data-dependency across the two channels (option attached to latest section), but I think this is not enforced in the code (as channel are polled in parallel, and select can pick in random order).

I think it will be better to embed the section-aware logic directly into the lexer, then having a single result channel.

unit/deserialize.go
```diff
@@ -43,36 +43,83 @@ var (
...
+// DeserializeSections deserializes into a list of UnitSections.
+func DeserializeSections(f io.Reader) ([]*UnitSection, error) {
+	sections, _, err := DeserializeAll(f)
+	return sections, err
+}
+
+// DeserializeAll deserializes into UnitSections and UnitOptions.
+func DeserializeAll(f io.Reader) ([]*UnitSection, []*UnitOption, error) {
+
+	lexer, secchan, optchan, errchan := newLexer(f)
```
lucab: (Here and everywhere else) Nit: I'd prefer naming those `sectCh` and `optCh`.

lucab:
@glawler thanks for the PR!
I left a few comments in review, including one major high-level concern with possible data-races with the new logic consuming lexer results.

unit/section.go
glawler: I tend to just put that on the assumption I will need imports at some point. I can remove it.

unit/serialize.go
glawler: Ack. Will fix.

unit/serialize.go
glawler: Will remove.

unit/deserialize.go
glawler: For that to happen the optchan would need to fire before the secchan. I don't believe that's possible given the code flow. All optchans fires are down code stream from secchan fires. 

I can certainly add a defensive check there though if you want.

unit/deserialize.go
glawler: I added a separate data structure as the original one did not fully support needed information. I added a new one rather than modified the existing one to support backwards compatibility. The new structure *does* contain *almost* the same information but also includes a bit of information that was not possible to express in the original structure, namely which options are in which sections when there are duplicate section names. 

I'm using this to support reading and writing `[Route]` sections into network interface unit files. The assumption of the current unit code seems to be that `section` names are unique. This is not true for `Route` sections as there can be multiple `Route` sections in a single file.

unit/deserialize.go
glawler: I was following the established code formats. I can change everywhere though, if you still want this.

unit/deserialize.go
glawler: I can do this, but think given the code flow, the dependency is ok. I'll take a look at making it a single channel though.

glawler:
I added an update to use a single channel as you suggested. Thanks.

unit/deserialize.go
lucab: Just for clarity, it doesn't matter the order in which they **fire**. What matters is the order in which `select` **receives** from them. As those are buffered channels, both can be ready to poll at the same time, in which case reception will be in random order (across the two channels).

FIFO order is guaranteed only on the single channel, i.e. each one separately.

unit/deserialize.go
```diff
@@ -43,37 +43,92 @@ var (
...
+		switch ld.Type {
+		case OPT:
+			if ld.Option != nil {
+				// add to options
+				opt := ld.Option
+				options = append(options, &(*opt))
+
+				// sanity check. "should not happen" as SEC is first in code flow.
+				if len(sections) == 0 {
+					// log error.
```
lucab: Feel free to actually return an hard error here.

unit/deserialize.go
lucab: Thanks. I do understand the underlying logic and goals. My doubts were about:
 * is there a usecase where you want to deserialize both in a single call? If not, this can be a private function.
 * balancing backward compatibility vs API usability, would it be better to just have `DeserializeOptions` and `DeserializeSections`?

glawler: Sorry, I misunderstood your comment entirely. Yes, this is just a helper function and I can think of no time you'd want both UnitSections and UnitOptions at once. This is just a utility function for  `Deserialize` and `DeserializeSections`. I will make this private. 

Are you suggesting renaming `Deserialize` to `DeserializeOptions`? Or writing a wrapper function and keeping `Deserialize`? Either way is OK with me, but keeping `Deserialize` would minimize disruption. Let me know which you prefer.

lucab: > Are you suggesting renaming Deserialize to DeserializeOptions?

I was thinking about this, yes. Not directly suggesting as I was not really sure whether I prefer compatibility over clarity.

I think in the end the best approach would be having:
 1) `DeserializeSections`
 2) `DeserializeOptions`
 3) `Deserialize` which forwards to `DeserializeOptions` and a deprecation notice in its docstring. Something like `this will be dropped in the future, use DeserializeOptions instead`.

unit/deserialize.go
```diff
@@ -43,18 +43,25 @@ var (
...
 
 // DeserializeSections deserializes into a list of UnitSections.
 func DeserializeSections(f io.Reader) ([]*UnitSection, error) {
-	sections, _, err := DeserializeAll(f)
+	sections, _, err := deserializeAll(f)
 	return sections, err
 }
 
+// Deserialize parses a systemd unit file into a list of UnitOptions.
+// Note: this function is depreciated in favor of DeserializeOptions
```
lucab: Typo: s/depreciated/deprecated/

lucab:
@glawler thanks! I spotted a minor typo, otherwise this seems ready to land. Please squash everything into a single commit.

PR #332, opened 2020-03-10
deserialize: Use quote to print possibly-invalid string

cgwalters:
An openshift/machine-config-operator user wrote an invalid unit,
and printing the contents as a byte array is unnecessary; it's
almost certain to be a valid string, we just found a syntax error.


cgwalters:
https://kubernetes.slack.com/archives/C6BRQSH2S/p1583847105180600

PR #335, opened 2020-06-02
docs: update docs links in README

lucab:
This updates documentation links in README, in order to use the new
module setup and abandon the now deprecated `godoc.org`.


lucab:
Docs update only, self-merging.

PR #336, opened 2020-06-02
dbus: add ListJobs method

ardias:
this adds method `org.freedesktop.systemd1.Manager.ListJobs`
returns all the jobs that are currently queued to be run

### DIFF
--- a/dbus/methods.go
+++ b/dbus/methods.go
@@ -598,3 +598,42 @@ func unitPath(name string) dbus.ObjectPath {
 func unitName(dpath dbus.ObjectPath) string {
 	return pathBusUnescape(path.Base(string(dpath)))
 }
+
+// Currently queued job definition
+type JobStatus struct {
+	Id       uint32          // The numeric job id
+	Unit     string          // The primary unit name for this job
+	JobType  string          // The job type as string
+	Status   string          // The job state as string
+	JobPath  dbus.ObjectPath // The job object path
+	UnitPath dbus.ObjectPath // The unit object path
+}
+
+// ListJobs returns an array with all currently queued jobs
+func (c *Conn) ListJobs() ([]JobStatus, error) {
+	return c.listJobsInternal()
+}
+
+func (c *Conn) listJobsInternal() ([]JobStatus, error) {
+	result := make([][]interface{}, 0)
+	if err := c.sysobj.Call("org.freedesktop.systemd1.Manager.ListJobs", 0).Store(&result); err != nil {
+		return nil, err
+	}
+
+	resultInterface := make([]interface{}, len(result))
+	for i := range result {
+		resultInterface[i] = result[i]
+	}
+
+	status := make([]JobStatus, len(result))
+	statusInterface := make([]interface{}, len(status))
+	for i := range status {
+		statusInterface[i] = &status[i]
+	}
+
+	if err := dbus.Store(resultInterface, statusInterface...); err != nil {
+		return nil, err
+	}
+
+	return status, nil
+}
--- a/dbus/methods_test.go
+++ b/dbus/methods_test.go
@@ -132,6 +132,19 @@ func runStopUnit(t *testing.T, conn *Conn, trTarget TrUnitProp) error {
 	return nil
 }
 
+func getJobStatusIfExists(jobs []JobStatus, jobName string) *JobStatus {
+	for _, j := range jobs {
+		if j.Unit == jobName {
+			return &j
+		}
+	}
+	return nil
+}
+
+func isJobStatusEmpty(job JobStatus) bool {
+	return job.Id == 0 && job.Unit == "" && job.JobType == "" && job.Status == "" && job.JobPath == "" && job.UnitPath == ""
+}
+
 // Ensure that basic unit starting and stopping works.
 func TestStartStopUnit(t *testing.T) {
 	target := "start-stop.service"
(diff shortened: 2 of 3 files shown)


dbus/methods.go
```diff
@@ -598,3 +598,43 @@ func unitPath(name string) dbus.ObjectPath {
 func unitName(dpath dbus.ObjectPath) string {
 	return pathBusUnescape(path.Base(string(dpath)))
 }
+
+type JobStatus struct {
```
lucab, line 602: As this is a public type with public fields, it would be nice to have docstrings everywhere.

dbus/methods.go
```diff
@@ -598,3 +598,43 @@ func unitPath(name string) dbus.ObjectPath {
...
+	Id       int
```
lucab, line 603: This is an uint32 according to the dbus method signature.

dbus/methods.go
```diff
@@ -598,3 +598,43 @@ func unitPath(name string) dbus.ObjectPath {
...
+	Unit     string
+	JobType  string
+	Status   string
+	JobPath  string
+	UnitPath string
```
lucab, line 608: This and the one above are `ObjectPath`.

dbus/methods.go
```diff
@@ -598,3 +598,43 @@ func unitPath(name string) dbus.ObjectPath {
...
+}
+
+// ListJobs returns an array with all currently queued jobs
+func (c *Conn) ListJobs() ([]JobStatus, error) {
+	return c.listJobsInternal(c.sysobj.Call("org.freedesktop.systemd1.Manager.ListJobs", 0).Store)
+}
+
+func (c *Conn) listJobsInternal(f storeFunc) ([]JobStatus, error) {
```
lucab, line 616: I think you copied this function-passing approach from other list methods in this module. I haven't touched that code in a while so I don't remember all the implications around that, but do you think it would be possible to avoid doing that here and just directly call the relevant method? I don't think there is much need for being more generic here.

dbus/methods.go
```diff
@@ -598,3 +598,43 @@ func unitPath(name string) dbus.ObjectPath {
...
+	result := make([][]interface{}, 0)
+	err := f(&result)
+	if err != nil {
```
lucab, line 619: I think you can merge this with the above call: `if err:= foo(); err != nil { .. }`

dbus/methods.go
```diff
@@ -598,3 +598,43 @@ func unitPath(name string) dbus.ObjectPath {
...
+	}
+
+	status := make([]JobStatus, len(result))
+	statusInterface := make([]interface{}, len(status))
+	for i := range status {
+		statusInterface[i] = &status[i]
+	}
+
+	err = dbus.Store(resultInterface, statusInterface...)
+	if err != nil {
```
lucab, line 635: Same comment as above.

dbus/methods_test.go
```diff
@@ -642,6 +651,49 @@ func TestEnableDisableUnit(t *testing.T) {
...
+
+	jobs, err := conn.ListJobs()
+	if err != nil {
+		t.Skip(err)
+	}
+
+	found := isJobPresent(jobs, service)
+	if !found {
+		t.Fatalf("%s job not found in list", service)
+	}
```
lucab, line 676: Can we also check the fields of the job we found here? Making sure that they aren't some dummy default value would be good already.

lucab:
@ardias thanks for the PR (and for writing the tests too)! I've left a few comments in in review, but overall it looks ok.

ardias:
@lucab Thanks for the feedback! I'll have the fixes in a few days

dbus/methods.go
brushknight: Done

dbus/methods.go
brushknight: Done

dbus/methods.go
brushknight: Done

dbus/methods.go
brushknight: Done

dbus/methods.go
brushknight: Done

dbus/methods.go
brushknight: Done

dbus/methods_test.go
brushknight: Done

dbus/methods_test.go
```diff
@@ -132,6 +132,19 @@ func runStopUnit(t *testing.T, conn *Conn, trTarget TrUnitProp) error {
...
+	for _, j := range jobs {
+		if j.Unit == jobName {
+			return &j
+		}
+	}
+	return nil
+}
+
+func isJobStatusEmpty(job JobStatus) bool {
+	return job.Id == 0 && job.Unit == "" && job.JobType == "" && job.Status == "" && job.JobPath == "" && job.UnitPath == ""
```
lucab, line 145: Nit: as unlikely as it could be, an `Id = 0` I think is a valid one which may result in test flakes, right? The rest of the condition seems fine on its own.

lucab:
@brushknight @ardias thanks both! I have a very minor nit for the test, otherwise this is good to go.

dbus/methods_test.go
brushknight: Just to clarify, we should remove check for Id, right? 
Also, should it be `&&` or `||` between rules?

lucab: Ah, I misread this check and already assumed the conditions where ORed. I think I agree with your current way of checking whether **all** fields are defaults. Just disregard my previous comment, sorry for the noise :)

brushknight:
@lucab can you tell us when this can be merged? Thank you in advance!

PR #343, opened 2020-10-02
daemon: Adds SdNotifyBarrier

rsms:
This adds daemon.SdNotifyBarrier including a unit test. It is modeled after [systemd's sd_notify_barrier](https://github.com/systemd/systemd/blob/v246/src/libsystemd/sd-daemon/sd-daemon.c#L556).

The barrier function is useful when a signalling process wants to wait for systemd to process previously-sent messages.

- I certify that I recognize and agree to the Developer Certificate of Origin of this project.
- Tested on go 1.15.2-linux-amd64 and 1.15.2-linux-386
- `go fmt` has been applied to the source code as that seems to be the case with other source files in this project.

Thanks for a useful package!

### DIFF
--- a/daemon/sdnotifybarrier.go
+++ b/daemon/sdnotifybarrier.go
@@ -33,6 +33,7 @@ var ErrEnvironment = errors.New("unsupported environment")
 // If `unsetEnvironment` is true, the environment variable `NOTIFY_SOCKET`
 // will be unconditionally unset.
 //
+// This feature was added in systemd v246
 func SdNotifyBarrier(ctx context.Context, unsetEnvironment bool) error {
 	// modelled after libsystemd's sd_notify_barrier
 
--- a/daemon/sdnotifybarrier.go
+++ b/daemon/sdnotifybarrier.go
@@ -1,0 +1,89 @@
+// Copyright 2020 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package daemon
+
+import (
+	"errors"
+	"io"
+	"net"
+	"os"
+	"syscall"
+	"time"
+)
+
+var ErrEnvironment = errors.New("unsupported environment")
+
+// SdNotifyBarrier allows the caller to synchronize against reception of
+// previously sent notification messages and uses the "BARRIER=1" command.
+//
+// If `unsetEnvironment` is true, the environment variable `NOTIFY_SOCKET`
+// will be unconditionally unset.
+//
+func SdNotifyBarrier(unsetEnvironment bool, timeout time.Duration) error {
+	// modelled after libsystemd's sd_notify_barrier
+
+	// construct unix socket address from systemd environment variable
+	socketAddr := &net.UnixAddr{
+		Name: os.Getenv("NOTIFY_SOCKET"),
+		Net:  "unixgram",
+	}
+	if socketAddr.Name == "" {
+		return ErrEnvironment
+	}
+
+	// create a pipe for communicating with systemd daemon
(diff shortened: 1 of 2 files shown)


rsms:
It appears an unrelated, unmodified test failed from timeout [according to Travis](https://travis-ci.org/github/coreos/go-systemd/jobs/732413698#L1720):

```
...
--- FAIL: TestPropertiesSubscription (10.01s)
1721    subscription_test.go:188: Reached timeout
...
```

daemon/sdnotifybarrier.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"errors"
+	"io"
+	"net"
+	"os"
+	"syscall"
+	"time"
+)
+
+var ErrEnvironment = errors.New("unsupported environment")
```
lucab, line 27: Missing docstring.

If possible, I'd prefer a less generic error description. Something along the line of "notification socket not available" and a similarly sounding variable name.

daemon/sdnotifybarrier.go
```diff
@@ -0,0 +1,89 @@
...
+
+// SdNotifyBarrier allows the caller to synchronize against reception of
+// previously sent notification messages and uses the "BARRIER=1" command.
+//
+// If `unsetEnvironment` is true, the environment variable `NOTIFY_SOCKET`
+// will be unconditionally unset.
+//
```
lucab, line 34: Spurious blank line.

This feature has been added quite recently in systemd `v246`, so it would be good to mention the minimum requirement here.

daemon/sdnotifybarrier.go
```diff
@@ -0,0 +1,89 @@
...
+func SdNotifyBarrier(unsetEnvironment bool, timeout time.Duration) error {
```
lucab, line 35: Just wondering, would it be possible to plumb a `context.Context` here instead of using an ad-hoc timeout? I guess that would also make it easier to use/understand the "no timeout" case.

lucab:
@rsms thanks for the patch and the accompanying test! I've left a few comments in review, the only major one about using a `context` for timing out. The CI sometimes flakes a bit, in which case you can force-push to re-trigger it.

daemon/sdnotifybarrier.go
rsms: Added a note about the systemd version.
The extra empty line at the end is a convention used by the go source code itself when a comment has multiple paragraphs. Since gofmt has no opinion on this it falls on you I guess, so I removed it :-)

daemon/sdnotifybarrier.go
```diff
@@ -0,0 +1,100 @@
...
+		return err
+	}
+	pipe_w.Close()
+
+	// wait for systemd to close the pipe
+	ctxch := make(chan struct{})
+	go func() {
+		select {
+		case <-ctx.Done():
+			pipe_r.SetReadDeadline(time.Now())
```
rsms, line 88: Note: `pipe_r.Close()` would work here too but setting an immediate read deadline has the upside of returning a descriptive error message: `i/o timeout`

daemon/sdnotifybarrier_test.go
```diff
@@ -0,0 +1,177 @@
...
+	"os"
+	"strings"
+	"syscall"
+	"testing"
+	"time"
+)
+
+func TestSdNotifyBarrier(t *testing.T) {
+
+	testDir, e := ioutil.TempDir("/tmp/", "test-")
```
rsms, line 31: Note: I used this since it's what I saw used in `TestSdNotify` and assumed there was a good reason (like a CI limitation.) If that's not the case, the testing package provides a nice TempDir() function that might be better for these tests (and more portable.) https://golang.org/pkg/testing/#T.TempDir

daemon/sdnotifybarrier.go
lucab: Ah, I guess I'm up to some learning today. Do you have examples/references of this? Does some godoc renderer rely on this?

lucab:
@rsms can you please rebase on latest master? CI moved to github actions in the meanwhile.

daemon/sdnotifybarrier_test.go
kolyshkin: Indeed we're using t.TempDir since commit 1f212c8 (part of PR #466).

daemon/sdnotifybarrier.go
kolyshkin: Better make this remark into an actual code comment.

kolyshkin:
@rsms I apologize the work to merge this PR is stalled. If you still want to work on this, can you please open a new PR (with all the commits from this one squashed and rebased on top of the current HEAD)?

PR #344, opened 2020-10-05
feat: add GetBootID support

hudeng-go:
Boot ID is needed sometimes


lucab:
Thanks for the PR! Can you please run your changes through `gofmt` and also add a test covering ` id, err := GetBootID() ==> err != nil && id != ""`?

lucab:
Visiting for triage, this still needs a rework.

PR #350, opened 2020-12-11
dbus: add GetUnitFileState method

vaspahomov:


dbus/methods.go
```diff
@@ -830,3 +830,21 @@ func (c *Conn) listJobsInternal(ctx context.Context) ([]JobStatus, error) {
...
+	err = obj.CallWithContext(ctx, "org.freedesktop.systemd1.Manager.GetUnitFileState", 0, name).Store(&prop)
+	if err != nil {
+		return "", err
+	}
+	defer func() {
+		if r := recover(); r != nil {
+			err = fmt.Errorf("failed to cast UnitFileState prop to string: %v", r)
+		}
+	}()
+	state := prop.Value().(string)
```
nekto0n, line 848: `state, ok :=...`?

dbus/methods.go
```diff
@@ -830,3 +830,21 @@ func (c *Conn) listJobsInternal(ctx context.Context) ([]JobStatus, error) {
...
+func (c *Conn) GetUnitFileStateContext(ctx context.Context, name string) (string, error) {
+	var err error
+	var prop dbus.Variant
+	obj := c.sysconn.Object("org.freedesktop.systemd1", "/org/freedesktop/systemd1")
+	err = obj.CallWithContext(ctx, "org.freedesktop.systemd1.Manager.GetUnitFileState", 0, name).Store(&prop)
+	if err != nil {
+		return "", err
+	}
+	defer func() {
+		if r := recover(); r != nil {
```
lucab, line 844: Let's get rid of this, and perform a checked type-assertion instead.

dbus/methods.go
```diff
@@ -830,3 +830,21 @@ func (c *Conn) listJobsInternal(ctx context.Context) ([]JobStatus, error) {
 
 	return status, nil
 }
+
+// GetUnitFileStateContext returns UnitFileState
+func (c *Conn) GetUnitFileStateContext(ctx context.Context, name string) (string, error) {
+	var err error
+	var prop dbus.Variant
+	obj := c.sysconn.Object("org.freedesktop.systemd1", "/org/freedesktop/systemd1")
+	err = obj.CallWithContext(ctx, "org.freedesktop.systemd1.Manager.GetUnitFileState", 0, name).Store(&prop)
```
lucab, line 839: You can declare `err` directly here.

lucab (changes requested):

lucab:
@vaspahomov thanks for your patch. I've left a few comments in review, and it would be good to have a test for this too.

dbus/methods_test.go
```diff
@@ -1600,3 +1602,27 @@ func TestUnitName(t *testing.T) {
 		}
 	}
 }
+
+// TestGetUnitFileState reads the `systemd-udevd.service` which should exist on all systemd
+// systems and ensures that UnitFileState property is valid.
+func TestGetUnitFileState(t *testing.T) {
+	conn := setupConn(t)
+	defer conn.Close()
+	service := "systemd-udevd.service"
```
lucab, line 1611: This may not exist in the testing environment. It would be good to set up a dedicated temporary service unit instead.

lucab (changes requested):
Code looks fine but CI is failing. I think this is because you are trying to re-use a system service unit that may not exist on older OS version. It would be good to try with a dedicated unit instead.

GaikwadPratik:
@vaspahomov , 

Can you please resolve conflicts so @lucab can merge it?

sigmonsays:
looking to call this method (GetUnitFileState) but it's just stale. Is there another way?

dbus/methods.go
```diff
@@ -830,3 +830,18 @@ func (c *Conn) listJobsInternal(ctx context.Context) ([]JobStatus, error) {
 
 	return status, nil
 }
+
+// GetUnitFileStateContext returns UnitFileState
```
kolyshkin, line 834: 1. You don't have to add `Context` to the method name -- all new methods accept context.
2. What is `UnitFileState` and where it is defined?

dbus/methods.go
```diff
@@ -830,3 +830,18 @@ func (c *Conn) listJobsInternal(ctx context.Context) ([]JobStatus, error) {
...
+func (c *Conn) GetUnitFileStateContext(ctx context.Context, name string) (string, error) {
+	var prop dbus.Variant
```
kolyshkin, line 836: Can you just use string here?

kolyshkin:
@vaspahomov please let us know if you want to keep working on this (and I apologize for a late response).

PR #354, opened 2021-02-25
sdjournal: add GetBootID support

hudeng-go:


hudeng-go:
@lucab this pull request add GetBootID test function.

sdjournal/journal.go
```diff
@@ -1118,3 +1136,26 @@ func (j *Journal) GetCatalog() (string, error) {
...
+		return "", err
+	}
+
+	var boot_id C.sd_id128_t
+	r := C.my_sd_id128_get_boot(sd_id128_get_boot, &boot_id)
+	if r < 0 {
+		return "", fmt.Errorf("failed to get boot id: %s", syscall.Errno(-r).Error())
+	}
+
+	sd_id128_to_string, err := getFunction("sd_id128_to_string")
```
lucab, line 1153: There seems to be a missing `err` check here.

sdjournal/journal.go
```diff
@@ -1118,3 +1136,26 @@ func (j *Journal) GetCatalog() (string, error) {
...
+	c := (*C.char)(C.malloc(32))
```
lucab, line 1154: I think this constant should be `33`, i.e. `SD_ID128_STRING_MAX`, in order to account for the NUL terminator.

sdjournal/journal.go
hudeng-go: It has been processed.

sdjournal/journal.go
```diff
@@ -1118,3 +1136,30 @@ func (j *Journal) GetCatalog() (string, error) {
...
+	}
+
+	sd_id128_to_string, err := getFunction("sd_id128_to_string")
+	if err != nil {
+		return "", err
+	}
+
+	c := (*C.char)(C.malloc(33))
+	defer C.free(unsafe.Pointer(c))
+	C.my_sd_id128_to_string(sd_id128_to_string, boot_id, c)
```
lucab, line 1160: This is discarding the return value and thus possibly ignoring error values. It should capture `r` and check it (like in the other C call above).

hudeng-go: my_sd_id128_to_string return is char, so I'm going to deal with possible errors by checking the bootid length.

lucab: Sorry, my bad, I mixed it with `_from_string()`.

lucab (approved):

lucab:
LGTM, thanks for the patch!

Issue #355, opened 2021-03-05
could not determine kind of name for C.my_sd_id128_to_string


learning813:
Hello,

I'm trying to compile a go source code and facing the below error while compiling . 

`# github.com/coreos/go-systemd/sdjournal
src/github.com/coreos/go-systemd/sdjournal/journal.go:1160:2: could not determine kind of name for C.my_sd_id128_to_string
cgo: 
gcc errors for preamble:
src/github.com/coreos/go-systemd/sdjournal/journal.go:313:60: error: '_SD_ARRAY_STATIC' undeclared here (not in a function)
 // my_sd_id128_to_string(void *f, sd_id128_t boot_id, char s[_SD_ARRAY_STATIC SD_ID128_STRING_MAX])
                                                            ^
In file included from /usr/include/systemd/sd-journal.h:31:0,
                 from src/github.com/coreos/go-systemd/sdjournal/journal.go:27:
src/github.com/coreos/go-systemd/sdjournal/journal.go:313:77: error: expected ']' before numeric constant
 // my_sd_id128_to_string(void *f, sd_id128_t boot_id, char s[_SD_ARRAY_STATIC SD_ID128_STRING_MAX])
                                                                             ^

error: Bad exit status from /var/tmp/rpm-tmp.sW2Hna (%build)

RPM build errors:
    Bad exit status from /var/tmp/rpm-tmp.sW2Hna (%build)
make: *** [rpm] Error 1`

My golang version is 1.13 and linux is centos 7

lucab:
I think this is because your version of libsystemd is missing the definition of `_SD_ARRAY_STATIC`. But I think it could be safely omitted from here. As we don't have CI for such ancient versions, it would be helpful if you could test whether https://github.com/coreos/go-systemd/pull/358 builds in your system.

PR #358, opened 2021-03-12
sdjournal: drop conditional static specifier

lucab:
`_SD_ARRAY_STATIC` is a custom / conditional macro for static C99
array size, which got introduced in recent libsystemd and thus cannot
build on older versions.
Due to how it is currently used in our Go wrapper, it can be safely
omitted from here.

Ref: systemd/systemd@4199f68
Fixes: https://github.com/coreos/go-systemd/issues/355


giuseppe (approved):
LGTM

it solves an issue we are seeing with Podman as well as the build on RHEL 8.x is currently failing

lucab:
@giuseppe this fix is now into https://github.com/coreos/go-systemd/releases/tag/v22.3.0.

PR #361, opened 2021-03-16
dbus: add KillUnitWithTarget

sargun:
This exposes the ability to send a kill signal to a specific
(sub)process in the unit, rather than all the processes that compose
the unit.

### DIFF
--- a/dbus/methods.go
+++ b/dbus/methods.go
@@ -24,6 +24,18 @@ import (
 	"github.com/godbus/dbus/v5"
 )
 
+// Who can be used to specify which process to kill in the unit via the KillUnitWithTarget API
+type Who string
+
+const (
+	// All sends the signal to all processes in the unit
+	All Who = "all"
+	// Main sends the signal to the main process of the unit
+	Main Who = "main"
+	// Control sends the signal to the control process of the unit
+	Control Who = "control"
+)
+
 func (c *Conn) jobComplete(signal *dbus.Signal) {
 	var id uint32
 	var job dbus.ObjectPath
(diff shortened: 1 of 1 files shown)


dbus/methods.go
```diff
@@ -196,7 +196,21 @@ func (c *Conn) KillUnit(name string, signal int32) {
 
 // KillUnitContext same as KillUnit with context
 func (c *Conn) KillUnitContext(ctx context.Context, name string, signal int32) {
-	c.sysobj.CallWithContext(ctx, "org.freedesktop.systemd1.Manager.KillUnit", 0, name, "all", signal).Store()
+	c.KillSpecificProcessUnitContext(ctx, name, All, signal)
+}
+
+type Who string
```
lucab, line 202: This (and the constants below) is a public entry in the API, and thus would need docstrings. Also, it should float to the top of the file.

dbus/methods.go
```diff
@@ -196,7 +196,21 @@ func (c *Conn) KillUnit(name string, signal int32) {
...
+
+const (
+	All     Who = "all"
+	Main    Who = "main"
+	Control Who = "control"
+)
+
+// KillSpecificProcessUnitContext is like KillUnitContext, but allows you to specify which process in the unit to send
+// the signal to.
+func (c *Conn) KillSpecificProcessUnitContext(ctx context.Context, name string, who Who, signal int32) error {
```
lucab, line 212: Bikeshedding: this a bit of a mouthful, and I think for new APIs we can avoid the `Context` suffix. Maybe `KillUnitWithTarget` and then rename the argument as `target Who`?

lucab:
Thanks for the PR! Happy to take this, I've left a few minor comments in review.

Unrelated, but overall this looks like a quite weird systemd API. I'm curious to read what is your real-world usecase for it, if you can share that.

sargun:
Sure --
The use case is to send an initial kill signal to a unit with RemainAfterExit=true to get it to stop, but without causing relevant state information to be deleted. This makes it so I do not need to lookup the mainpid and subsequently send a signal.

lucab (approved):

PR #362, opened 2021-03-24
sdjournal: fix input argument type in C.Malloc call

lsm5:
podman v3.1.0-rc2 uses v22 go-systemd and fedora rpm builds were failing
for 32-bit arches (i686 and armv7hl).

This commit conditionally defines id128StringMax based on
32-bit v/s 64-bit arches.

Signed-off-by: Lokesh Mandvekar <lsm5@fedoraproject.org>
Closes: https://github.com/coreos/go-systemd/issues/363

/cc @vrothberg @ashcrow @mheon @rhatdan

### DIFF
--- a/sdjournal/journal.go
+++ b/sdjournal/journal.go
@@ -1155,7 +1155,7 @@ func (j *Journal) GetBootID() (string, error) {
 		return "", err
 	}
 
-	id128StringMax := C.ulong(C.SD_ID128_STRING_MAX)
+	id128StringMax := C.size_t(C.SD_ID128_STRING_MAX)
 	c := (*C.char)(C.malloc(id128StringMax))
 	defer C.free(unsafe.Pointer(c))
 	C.my_sd_id128_to_string(sd_id128_to_string, boot_id, c)


sdjournal/journal_64.go
```diff
@@ -0,0 +1,33 @@
+// +build !i386 !arm
```
mheon, line 1: This should probably be `!i386,!arm`

lsm5: ack, yes, changing.

lucab (changes requested):
Thanks for the PR, but a bug report beforehand with logs would have been much appreciated.
I'm very doubtful of this patch, can you please share the actual build failure on whatever 32b platform you have at hand?

From a quick skim, I guess the actual bug is that the cast-target type should have been `C.size_t` instead.

lsm5:
> Thanks for the PR, but a bug report beforehand with logs would have been much appreciated.
> I'm very doubtful of this patch, can you please share the actual build failure on whatever 32b platform you have at hand?

I'll regenerate the build failure log

lsm5:
filed https://github.com/coreos/go-systemd/issues/363

sdjournal/journal_64.go
```diff
@@ -0,0 +1,33 @@
+// +build !i386,!arm
+
+// Copyright 2015 RedHat, Inc.
```
rhatdan, line 3: If this is new code it should only by Red Hat and 2021.

sdjournal/journal.go
```diff
@@ -1155,7 +1155,6 @@ func (j *Journal) GetBootID() (string, error) {
 		return "", err
 	}
 
-	id128StringMax := C.ulong(C.SD_ID128_STRING_MAX)
```
lucab, line 1158:
```
id128StringMax := C.size_t(C.SD_ID128_STRING_MAX)
```

should be enough to fix this everywhere. Can you try a scratch build on 32b just to confirm, and if so amend the patch?

lucab:
> cannot use id128StringMax (type _Ctype_ulong) as type _Ctype_uint in argument to _Cfunc__CMalloc

Indeed this hints at a type mismatch in `C.Malloc` argument, which should be a `C.size_t`. See suggestion for a better patch in the review comment.

sdjournal/journal.go
lsm5: that worked.

lsm5:
> > cannot use id128StringMax (type _Ctype_ulong) as type _Ctype_uint in argument to _Cfunc__CMalloc
> 
> Indeed this hints at a type mismatch in `C.Malloc` argument, which should be a `C.size_t`. See suggestion for a better patch in the review comment.

Yes, patch updated. Thanks a lot @lucab !!

lucab (approved):

lsm5:
@lucab thanks a lot! I think we'd need a new release with this change /cc @mheon.

mheon:
Would be appreciated, though if not possible we can revert to an earlier release.

lucab:
Tagged: https://github.com/coreos/go-systemd/releases/tag/v22.3.1

mheon:
Greatly appreciated!

PR #367, opened 2021-04-14
journal: close unix socket after successful probe

### LINKED ISSUE #366: journald unix sockets are leaked whenever Enabled() is called
The journald logging code appears to leak a unix datagram socket once per call to `journal.Enabled()` on linux platforms.

An example bit of code follows that leads to an eventual error once the number of open file handles hits the operating system per-process maximum.

```
package main

import (
	"fmt"

	"github.com/rs/zerolog"
	"github.com/rs/zerolog/journald"
	"github.com/rs/zerolog/log"
)

func main() {

	var (
		i int
	)

	writers := journald.NewJournalDWriter()
	log.Logger = log.Output(writers)

	for {
		log.WithLevel(zerolog.InfoLevel).Msg(fmt.Sprintf("Hello world %d", i))
		i++
	}
}
```

jeremy-gill:
This is done to ensure unix sockets aren't left dangling after each Enabled() call, which
can lead to eventual process resource exhaustion. Fixes #366.

### DIFF
--- a/journal/journal_unix.go
+++ b/journal/journal_unix.go
@@ -65,9 +65,11 @@ func Enabled() bool {
 		return false
 	}
 
-	if _, err := net.Dial("unixgram", journalSocket); err != nil {
+	conn, err := net.Dial("unixgram", journalSocket)
+	if err != nil {
 		return false
 	}
+	defer conn.Close()
 
 	return true
 }


lucab:
Good catch, and thanks for the separate bug report too! It is my understanding that the garbage collector would eventually take care of the stale socket, and that most consumers would do the probe only once upfront and then cache the result.
However I totally see how this could hit the limit of open FDs in a tight loop, and closing the socket when done is a good practice anyway.

journal/journal_unix.go
```diff
@@ -65,10 +65,14 @@ func Enabled() bool {
 		return false
 	}
 
-	if _, err := net.Dial("unixgram", journalSocket); err != nil {
+	conn, err := net.Dial("unixgram", journalSocket)
+
```
lucab, line 69: Spurious new line.

journal/journal_unix.go
```diff
@@ -65,10 +65,14 @@ func Enabled() bool {
...
+	if err != nil {
 		return false
 	}
 
+	defer conn.Close()
```
lucab, line 74: Let move this one line up, no need to have an extra empty line after the if-err body.

lucab (commented):
A couple of minor comments on spurious new lines. When fixing those, can you please squash everything into a single final commit?

journal/journal_unix.go
jeremy-gill: I've squashed the two commits into one and removed spurious blank lines.  Does this need a unit test?

jeremy-gill (commented):
The consumer in question that triggered this issue is [zerolog](https://github.com/rs/zerolog).  Zerolog's journald backend writer checks journald status for each write (see [here](https://github.com/rs/zerolog/blob/master/journald/journald.go#L72)).  Sockets quickly pile up.  Perhaps there's a better way to ensure the connection is still alive in zerolog.

journal/journal_unix.go
jeremy-gill: See 68a12a3.

journal/journal_unix.go
jeremy-gill: See 68a12a3.

lucab (approved):

PR #368, opened 2021-04-25
journal: add string method to journal.Priority

bcicen:
Adds string method to journal Priority level

### DIFF
--- a/journal/journal.go
+++ b/journal/journal.go
@@ -40,6 +40,13 @@ const (
 	PriDebug
 )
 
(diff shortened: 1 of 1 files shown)


lucab:
Thanks for the report!
Do you maybe have a reference to the consumer where you need this?
I'm somehow inclined towards not merging this: those constants are mostly mirroring https://golang.org/pkg/log/syslog/#Priority (minus the facility part) and as such they don't have a canonical string representation (which is locale dependent too).

PR #373, opened 2021-06-10
dbus: implement support for cgroup freezer APIs

msekletar:


lucab:
Thanks for the PR.
There is a minimum-version breakage in tests, which I think can be directly massaged here.
Also, I think those are quite recent, so I think our CI here does not cover them yet.
It may be good to add a note in the docstrings about systemd version and cgroup requirements.

msekletar:
I don't think that we should be mentioning systemd version where these APIs where first implemented. Features can (and do get) backported to older versiond and hence our docs would be misleading.

msekletar:
Updated. @lucab PTAL.

lucab (approved):
LGTM, thanks!

PR #378, opened 2021-08-31
dbus: support get underlying sysconn and sigconn

ImSingee:
Sometimes we may want to get the underlying dbus connection's status.

A simple `Connection` method is introduced for re-using the connection 

```go
var globalConn = dbus.NewSystemdConnectionContext(ctx)

func getConn() *Conn {
   if globalConn.Connected() {
      // refresh the connection
      globalConn = dbus.NewSystemdConnectionContext(ctx)
   }

   return globalConn
}

func main() {
     getConn().xxx()

    // between the two use of the getConn, somethings may happen to break the connection
    // e.g.: systemctl daemon-reexec

    getConn().xxx()
}

```

And the underlying `sysconn` and `sigconn` are also exported in this PR. A typical usage is in Subscription

```
func main() {
    for {
        conn := dbus.NewSystemdConnectionContext(context.Background())

        conn.Subscribe()
        sets := conn.NewSubscriptionSet()
        changes, errs := sets.Subscribe()

        for {
          select {
          case change := <-changes:
              // ...
          case err := <-errs:
             // ...
          case <- conn.SigConn().Context().Done():
             continue
          case <- conn.SysConn().Context().Done():
             continue
          }
      }
    }

}

```

### DIFF
--- a/dbus/methods.go
+++ b/dbus/methods.go
@@ -417,6 +417,29 @@ func (c *Conn) listUnitsInternal(f storeFunc) ([]UnitStatus, error) {
 	return status, nil
 }
 
+// GetUnitByPID returns the unit object path of the unit a process ID
+// belongs to. It takes a UNIX PID and returns the object path. The PID must
+// refer to an existing system process
+func (c *Conn) GetUnitByPID(ctx context.Context, pid uint32) (dbus.ObjectPath, error) {
+	var result dbus.ObjectPath
+
+	err := c.sysobj.CallWithContext(ctx, "org.freedesktop.systemd1.Manager.GetUnitByPID", 0, pid).Store(&result)
+
+	return result, err
+}
+
+// GetUnitNameByPID returns the name of the unit a process ID belongs to. It
+// takes a UNIX PID and returns the object path. The PID must refer to an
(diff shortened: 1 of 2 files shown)


lucab:
@ImSingee Thanks for the PR!
It looks like the problem you are facing is due to not having an ergonomic done/cancellation channel. I'm not sure exposing `sysconn` and `sigconn` is the right way to go though.
Internally we already have contexts on the relevant entities. Thus I think it would be better to properly chain all the contexts to the top-level one, and arranging things so that a cancellation from any of the inferior ones gets forwarded up. Does that sound reasonable to you?

ImSingee:
Sorry, but the scenario is when the dbus connection is disconnected at the system level (e.g. systemctl daemon-rexec) rather than at the program level, especially for subscribing systemd properties change, and we need a mechanism to know that the dbus service restart (and lead to our existing sigconn disconnected) and we should reconnect.
Since this disconnection happens outside the program, we have no way to monitor it through a context of our own, so we do need the context of underlying sigconn object

invidian:
This PR has 0 commits, I guess it can be closed?

PR #379, opened 2021-09-20
dbus: add GetUnitProcessesContext to list any unit's running processes 

pphysch:
This PR adds a binding for `org.freedesktop.systemd1.Manager.GetUnitProcesses`, taking a unit string and returning a list of all processes currently running under that unit (including child units).

This is convenient for getting a lot of information about a Slice's running processes, as it descends into child Slices/Scopes and includes those processes in the list. For example, if we target `user-1000.slice` we can learn about the different `session-*` Scopes under that user and all the processes therein.

My test (which may need adjustments to be more idiomatic/isolated) passes on systemd v239 (RHEL 8), but this Manager method doesn't exist on v219 (RHEL 7). I dug for a bit and can't find the exact version that implemented the GetProcesses API, but it seems to have been around 5 years ago.

This is my first PR here so please let me know how I can fix it! I used existing Manager method bindings as a template, so the code itself should be conformant even if the test needs work.

### DIFF
--- a/dbus/methods.go
+++ b/dbus/methods.go
@@ -841,12 +841,12 @@ func (c *Conn) ThawUnit(ctx context.Context, unit string) error {
 }
 
 type Process struct {
-	Path    string // The primary unit name as string
-	PID     uint64
-	Command string
+	Path    string // Where this process exists in the unit/cgroup hierarchy
+	PID     uint64 // The numeric process ID (PID)
+	Command string // The process command and arguments as a string
 }
 
-// GetUnitProcessesContext returns an array with all currently running processes in a unit
+// GetUnitProcessesContext returns an array with all currently running processes in a unit *including* its child units.
 func (c *Conn) GetUnitProcessesContext(ctx context.Context, unit string) ([]Process, error) {
 	return c.getUnitProcessesInternal(ctx, unit)
 }
--- a/dbus/methods.go
+++ b/dbus/methods.go
@@ -839,3 +839,38 @@ func (c *Conn) FreezeUnit(ctx context.Context, unit string) error {
 func (c *Conn) ThawUnit(ctx context.Context, unit string) error {
 	return c.sysobj.CallWithContext(ctx, "org.freedesktop.systemd1.Manager.ThawUnit", 0, unit).Store()
 }
+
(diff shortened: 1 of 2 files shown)


dbus/methods_test.go
```diff
@@ -1657,3 +1659,41 @@ func TestFreezer(t *testing.T) {
...
+	exists := false
+	for _, p := range processes {
+		if p.PID == pid && strings.HasPrefix(p.Command, "/bin/sleep") {
+			exists = true
+			t.Logf("Found %v\n", p)
+		}
+	}
+
+	if !exists {
+		t.Errorf("PID %d ('/bin/sleep 400') not found in current Slice unit's process list", pid)
```
lucab, line 1697: It looks like this test is failing on some CI runs. I suspect there is a race between spawning the child `sleep` and getting unit processes. It could be useful to retry the `GetUnitProcessesContext()` a few times until we get a non-empty set with `/bin/sleep`, or eventually timeout if we never get that.

dbus/methods.go
```diff
@@ -839,3 +839,38 @@ func (c *Conn) FreezeUnit(ctx context.Context, unit string) error {
...
 }
+
+type Process struct {
+	Path    string // Where this process exists in the unit/cgroup hierarchy
+	PID     uint64 // The numeric process ID (PID)
+	Command string // The process command and arguments as a string
+}
+
+// GetUnitProcessesContext returns an array with all currently running processes in a unit *including* its child units.
+func (c *Conn) GetUnitProcessesContext(ctx context.Context, unit string) ([]Process, error) {
```
lucab, line 850: I think we can directly call this `GetUnitProcesses`, plus avoid the separate `*Internal` helper.

lucab:
Thanks for the PR! It looks like the new test is frequently failing in CI, I think because it is internally racing (see review).

pphysch:
Looks like the test is still failing even with delayed retries; I'll rewrite it so that it does more mocking and has less reliance on the system/container.

pphysch:
Any chance this PR can get reviewed? Tests were passing as of the last commit.

kolyshkin (changes requested):

PR #385, opened 2021-12-13
dbus: add support for querying unit by PID

dylanratcliffe:
This exposes the functionality of the `GetUnitByPID` method and could be useful in situations where you're trying to find which unit owns a given process.

I chose to convert from `dbus.ObjectPath` to a `string` (the name of the unit) since almost nothing in this library accepts an `dbus.ObjectPath` as an input, and the methods for converting back and forth are private and not available to users. The current output format could then be passed into `ListUnitsByNames` in order to get further information.

### DIFF
--- a/dbus/methods.go
+++ b/dbus/methods.go
@@ -417,6 +417,24 @@ func (c *Conn) listUnitsInternal(f storeFunc) ([]UnitStatus, error) {
 	return status, nil
 }
 
+func (c *Conn) getUnitInternal(f storeFunc) (string, error) {
+	var result dbus.ObjectPath
+
+	err := f(&result)
+
+	// Nothing in this library actually accepts a dbus.ObjectPath, so it's much
+	// more useful as a name
+	name := unitName(result)
+
+	return name, err
+}
+
+// GetUnitByPIDContext returns the unit name for a given PID. The PID must refer
+// to an existing system process
+func (c *Conn) GetUnitByPIDContext(ctx context.Context, pid uint32) (string, error) {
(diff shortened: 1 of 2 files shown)


dbus/methods.go
```diff
@@ -417,6 +417,24 @@ func (c *Conn) listUnitsInternal(f storeFunc) ([]UnitStatus, error) {
...
+	// Nothing in this library actually accepts a dbus.ObjectPath, so it's much
+	// more useful as a name
+	name := unitName(result)
+
+	return name, err
+}
+
+// GetUnitByPIDContext returns the unit name for a given PID. The PID must refer
+// to an existing system process
+func (c *Conn) GetUnitByPIDContext(ctx context.Context, pid uint32) (string, error) {
```
lucab, line 434: We can drop the `Context` suffix from here (it's an historical artifact added for method that started their existence before `context` was a thing).

I agree getting the unit name is probably the most common case, but the object path could be useful too in some case.
I'd suggest having a plain `GetUnitByPID` that directly returns the object path, and an additional `GetUnitNameByPID` which internally does some translation. No need for a separate `getUnitInternal` helper, that logic can be folded directly into the latter.

lucab:
Welcome and thanks for the PR! Overall it looks nice, I've left some minor suggestions in review.

dbus/methods.go
dylanratcliffe: Works for me, should be addressed

lucab (approved):
LGTM, and thanks for adding test coverage too!

Issue #389, opened 2022-01-10
Rename login1.Conn to login1.Client or similar


invidian:
I think `login1.Conn` name is not very accurate, as it leaks the abstraction on how login1 methods are called underneath (via some kind of connection).

Also, right now login1 package is responsible for connecting and authenticating to D-Bus, which seems like both possible waste of resources, as multiple clients can't share the connection and it also affects testability of code, as no mock D-Bus connection can be given to a login1 client. It seems like responsibility to proxy connections to D-Bus should be out of scope for this package/struct.

PR #390, opened 2022-01-10
login1: add NewWithConnection and RebootWithContext methods

### LINKED ISSUE #387: Provide error handling for Reboot() call
Right now user has no way of knowing if calling `Reboot()` succeeds or not, as no error is returned from a call. E.g. if user calling it has no permissions to do so, the call will silently fail.

This could be improved, e.g. together with accepting the context for operation in new function.

invidian:
This PR does some small improvements to `login1` package, namely changes tests to use blackbox approach to avoid building up the tech debt by adding new tests without doing so in the first place.

Then adds `NewWithConnection` method together with tests as a new constructor to login1.Conn (#389), which allows re-using and mocking D-Bus connection so remaining method calls can be tested.

Finally, `RebootWithContext` method is added as an replacement and improved version of `Reboot`.

Closes #387


lucab (changes requested):
Thanks for the PR!
This is unfortunately becoming a kitchen sink, which isn't nice.
Let's only add `RebootWithContext` in here, everything else will be touched elsewhere.

invidian:
Hey, thanks for quick feedback. Would you like me to create one PR per commit from this PR then? Generally, I'd prefer to not submit code without tests, no matter how trivial it is, and changes here as the simplest steps to get things testable. Also, I'm concerned that "everything else" will not get enough attention once `RebootWithContext` is merged, so I'd prefer to focus on other bits first.

lucab:
Generally speaking yes, I prefer handling separate PRs as they can easily travel at different speeds and merge at different times.
So I'd use this only to fix #387, which is indeed a very useful thing (and thanks again!).

Regarding testing, it usually goes in the same PR, agreed. In this case though it is both a large reworking and pretty much just formalized mocking. It effectively does not cover much juicy content, other than package-global constants and other static strings.
From a maintenance point of view, the increased ballast is not worth the price. Let's aggressively trim down this PR to just carry `RebootWithContext`.

invidian:
I've created https://github.com/coreos/go-systemd/pull/395 and https://github.com/coreos/go-systemd/pull/396 then, to split the changes and make them easier to review and merge.

For the time being solving #391 is not needed, I'll rework this PR once PRs above are merged, so it's not needed, so this one can be merged as well.

PR #393, opened 2022-01-17
dbus: add Connected methods to check connections status

wzshiming:
xref https://github.com/kubernetes/kubernetes/issues/100328#issuecomment-1014204604

Pass the Dbus Connected method


wzshiming:
Hi @lucab
Could you have time to take a look? 
this method is needed in the k8s/runc project.

lucab (approved):

PR #395, opened 2022-01-31
login1: Use blackbox testing

invidian:
This is a desired way of testing to avoid creating fragile test suites
and be able to refactor code without touching tests.

Extracted from #390


kolyshkin:
this is being carried in #503

PR #396, opened 2022-01-31
login1: Add NewWithConnection method 

invidian:
This method allows passing existing D-Bus connection, which allows to
re-use connection between clients and to mock D-Bus connection for
testing purposes.

Extracted from #390
Includes commits from #395


invidian:
Closing in favor of #503.

PR #397, opened 2022-02-01
journal: remove implicit initialization

drakkan:
some application could log to journald conditionally, the implicit
initialization will always create a socket on init.

The Enabled method can be used to initialize the logger and many
applications already use it.

### DIFF
--- a/journal/journal_unix.go
+++ b/journal/journal_unix.go
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build !windows
 // +build !windows
 
 // Package journal provides write bindings to the local systemd journal.
@@ -53,15 +54,9 @@ var (
 	onceConn sync.Once
 )
 
-func init() {
-	onceConn.Do(initConn)
-}
-
 // Enabled checks whether the local systemd journal is available for logging.
 func Enabled() bool {
-	onceConn.Do(initConn)
-
-	if (*net.UnixConn)(atomic.LoadPointer(&unixConnPtr)) == nil {
+	if c := getOrInitConn(); c == nil {
 		return false
 	}
 
@@ -82,7 +77,7 @@ func Enabled() bool {
 // (http://www.freedesktop.org/software/systemd/man/systemd.journal-fields.html)
 // for more details.  vars may be nil.
 func Send(message string, priority Priority, vars map[string]string) error {
-	conn := (*net.UnixConn)(atomic.LoadPointer(&unixConnPtr))
+	conn := getOrInitConn()
 	if conn == nil {
 		return errors.New("could not initialize socket to journald")
 	}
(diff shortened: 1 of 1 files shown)


drakkan:
uhmm... sdjournal test seems to depends on the implicit initialization. I can fix them but before I would like to know if this patch can be accepted in this way or with some modifications, thanks

lucab:
Thanks for the PR.
I think this is indeed worth exploring, but your current patch would break all existing consumers, which isn't really feasible.

But I think we can investigate adding the `Once.Do()` to the `Send()` codepath too.
That would keep existing consumers working, but would incur some runtime penalty in the hotpath, mmh.

Perhaps we can tweak both `Enabled()` and `Send()` in the following way:
 * perform an atomic load for `conn`; if non-nil we are good
 * if `nil`:
   - perform the `Once.Do(initConn)` 
   - try once again an atomic load for `conn`; if non-nil we are good
   - if still `nil`, it's a permanent failure, halt here and return

drakkan:
Glad you are considering exploring this contribution, thank you!

The patch should be backward compatible now, please advise if you want any other changes

journal/journal_unix.go
```diff
@@ -126,6 +121,16 @@ func Send(message string, priority Priority, vars map[string]string) error {
 	return nil
 }
 
+// getConn attempts to get the global `unixConnPtr` socket, initializing if necessary
+func getConn() *net.UnixConn {
```
lucab, line 125: I guess `getOrInitConn` may be a better name, perhaps?

The docstring on `initConn` should also be updated to reflect the new situation.

lucab:
@drakkan thanks, that looks better. I've left some minor comment in review. When done with it, please squash back the additional changes into the single initial commit.

journal/journal_unix.go
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build !windows
```
lucab, line 15: Uhm, this file is already called `_unix.go` so I think this (and the existing annotation just below too) may not be needed. Did you see something which suggests otherwise?

lucab: Ah bummer, I didn't realized `unix` and `linux` behave differently in regard to this. Thanks for the reference!

lucab (approved):
LGTM, thanks for the patch!

drakkan:
Thank you! Only for info, do you plan to tag a new release anytime soon?

lucab:
Bit of a busy period right now, but yes I hope to have a tag soonish. I think you can pick this up with a gomod `replace` in the meanwhile.

drakkan:
> Bit of a busy period right now, but yes I hope to have a tag soonish. I think you can pick this up with a gomod `replace` in the meanwhile.

I'll try to propose a patch to zerolog to use the latest version together with another fix and in the meantime I'll use a replace, thank you

PR #404, opened 2022-06-29
Add package for working with systemd credentials

### LINKED ISSUE #402: Module for reading service credentials
I would like a module that can read [credentials](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#Credentials) when the current program is running as a systemd service.

This should be fairly simple and a small amount of code. I'm willing to send a PR at some point.

cedws:
Closes #402.

Criticism is welcome, this is just some quick code I wrote recently and got permission from my employer to contribute.


cedws:
Oh, `io/fs` is not available in Go 1.15. I'll get around to doing it some other way unless we can bump the CI version.

PR #408, opened 2022-08-30
login1: add context-aware ListSessions and ListUsers methods

ilyam8:
Adding those two because I need only them at the moment. The change is super straightforward (Call => CallWithContext).


lucab (approved):
LGTM, thanks for the patch!

PR #409, opened 2022-08-30
login1: add methods to get session/user properties

ilyam8:
This PR adds context-aware methods to get a property/all properties for a user or session.

~~I am not sure how to unit-test them, the majority of login1 connection methods have no unit tests. I think it will be possible after #396 (will be possible to mock both dbus.Conn and dbus.BusObject)~~.

### DIFF
--- a/login1/dbus.go
+++ b/login1/dbus.go
@@ -26,9 +26,11 @@ import (
 )
 
 const (
-	dbusDest      = "org.freedesktop.login1"
-	dbusInterface = "org.freedesktop.login1.Manager"
-	dbusPath      = "/org/freedesktop/login1"
+	dbusDest             = "org.freedesktop.login1"
+	dbusManagerInterface = "org.freedesktop.login1.Manager"
+	dbusSessionInterface = "org.freedesktop.login1.Session"
+	dbusUserInterface    = "org.freedesktop.login1.User"
+	dbusPath             = "/org/freedesktop/login1"
 )
 
 // Conn is a connection to systemds dbus endpoint.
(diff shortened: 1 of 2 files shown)


login1/dbus.go
```diff
@@ -25,9 +26,11 @@ import (
 )
 
 const (
-	dbusDest      = "org.freedesktop.login1"
-	dbusInterface = "org.freedesktop.login1.Manager"
-	dbusPath      = "/org/freedesktop/login1"
+	dbusDest             = "org.freedesktop.login1"
+	dbusInterface        = "org.freedesktop.login1.Manager"
```
lucab, line 30: Let's call this `dbusManagerInterface` then maybe?

login1/dbus.go
```diff
@@ -341,3 +364,35 @@ func (c *Conn) Subscribe(members ...string) chan *dbus.Signal {
...
+	var props map[string]dbus.Variant
+	err := obj.CallWithContext(ctx, "org.freedesktop.DBus.Properties.GetAll", 0, dbusInterface).Store(&props)
+	if err != nil {
+		return nil, err
+	}
+
+	return props, nil
+}
+
+func (c *Conn) getProperty(ctx context.Context, path dbus.ObjectPath, dbusInterface, property string) (dbus.Variant, error) {
```
lucab, line 384: I wonder if we should return the resulting `Variant` by reference instead, maybe?.

lucab:
Thanks for the patch! I left some comments in review. Feel free to squash the commits when you think you are done, and I'll have a further look.

lucab:
Indeed, this logic doesn't have much unit-testing coverage, as that really doesn't add much.
The test suite does a bit of integration testing, which is harder to balance though.
I think the tests that you added as sanity checks do make sense.

ilyam8:
@lucab, thanks for the quick review 🙋‍♂️ I believe I've addressed all your comments.

lucab (approved):
LGTM, thanks for the patch!

PR #410, opened 2022-10-19
journal: add StderrIsJournalStream function

WGH-:
This function can be used for automatic protocol upgrade described in [1].

Both unit tests and runnable example are included. Only the latter requires systemd, as unit tests are self-sufficient, and only test that JOURNAL_STREAM environment variable is checked properly.

[1] https://systemd.io/JOURNAL_NATIVE_PROTOCOL/#automatic-protocol-upgrading

### DIFF
--- a/journal/journal_unix.go
+++ b/journal/journal_unix.go
@@ -82,6 +82,24 @@ func Enabled() bool {
 //
 // [Journal Native Protocol]: https://systemd.io/JOURNAL_NATIVE_PROTOCOL/#automatic-protocol-upgrading
 func StderrIsJournalStream() (bool, error) {
+	return fdIsJournalStream(syscall.Stderr)
+}
+
+// StdoutIsJournalStream returns whether the process stdout is connected
+// to the Journal's stream transport.
+//
+// Returns true if JOURNAL_STREAM environment variable is present,
+// and stdout's device and inode numbers match it.
+//
+// Error is returned if unexpected error occurs: e.g. if JOURNAL_STREAM environment variable
+// is present, but malformed, fstat syscall fails, etc.
+//
+// Most users should probably use [StderrIsJournalStream].
+func StdoutIsJournalStream() (bool, error) {
+	return fdIsJournalStream(syscall.Stdout)
+}
+
+func fdIsJournalStream(fd int) (bool, error) {
 	journalStream := os.Getenv("JOURNAL_STREAM")
 	if journalStream == "" {
 		return false, nil
@@ -94,7 +112,7 @@ func StderrIsJournalStream() (bool, error) {
 	}
 
 	var stat syscall.Stat_t
-	err = syscall.Fstat(syscall.Stderr, &stat)
+	err = syscall.Fstat(fd, &stat)
 	if err != nil {
 		return false, err
 	}
--- a/journal/journal_unix.go
+++ b/journal/journal_unix.go
@@ -69,6 +69,40 @@ func Enabled() bool {
 	return true
 }
 
+// StderrIsJournalStream returns whether the process stderr is connected
+// to the Journal's stream transport.
+//
+// This can be used for automatic protocol upgrading described in [Journal Native Protocol].
+//
+// Returns true if JOURNAL_STREAM environment variable is present,
+// and stderr's device and inode numbers match it.
+//
+// Error is returned if unexpected error occurs: e.g. if JOURNAL_STREAM environment variable
+// is present, but malformed, fstat syscall fails, etc.
+//
+// [Journal Native Protocol]: https://systemd.io/JOURNAL_NATIVE_PROTOCOL/#automatic-protocol-upgrading
+func StderrIsJournalStream() (bool, error) {
+	journalStream := os.Getenv("JOURNAL_STREAM")
+	if journalStream == "" {
+		return false, nil
+	}
+
+	var expectedStat syscall.Stat_t
+	_, err := fmt.Sscanf(journalStream, "%d:%d", &expectedStat.Dev, &expectedStat.Ino)
+	if err != nil {
+		return false, fmt.Errorf("failed to parse JOURNAL_STREAM=%q: %v", journalStream, err)
+	}
+
+	var stat syscall.Stat_t
+	err = syscall.Fstat(syscall.Stderr, &stat)
+	if err != nil {
+		return false, err
+	}
+
+	match := stat.Dev == expectedStat.Dev && stat.Ino == expectedStat.Ino
+	return match, nil
+}
+
 // Send a message to the local systemd journal. vars is a map of journald
 // fields to values.  Fields must be composed of uppercase letters, numbers,
 // and underscores, but must not start with an underscore. Within these
--- a/journal/journal_windows.go
+++ b/journal/journal_windows.go
@@ -33,3 +33,7 @@ func Enabled() bool {
 func Send(message string, priority Priority, vars map[string]string) error {
 	return errors.New("could not initialize socket to journald")
 }
+
+func StderrIsJournalStream() (bool, error) {
+	return false, nil
+}
--- a/journal/journal_windows.go
+++ b/journal/journal_windows.go
@@ -37,3 +37,7 @@ func Send(message string, priority Priority, vars map[string]string) error {
 func StderrIsJournalStream() (bool, error) {
 	return false, nil
 }
+
+func StdoutIsJournalStream() (bool, error) {
+	return false, nil
+}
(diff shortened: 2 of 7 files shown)


lucab:
Thanks for the PR!

At the high level, I think this API should be renamed to avoid mentioning `stderr`. In fact, the upgrade dance works the same way for `stderr` as well as for `stdout`, it's just that the systemd documentation is slightly deceiving on that front. See https://github.com/systemd/systemd/issues/6800 for a concrete example of what I'm talking about.

In the Rust land this logic is implemented by https://docs.rs/libsystemd/0.5.0/libsystemd/logging/fn.connected_to_journal.html, so I think similarly here it could be called `IsConnectedToJournal()`. Also, it should check device and inode matching against `stdout` too (the final result is the `OR` of the two checks).

WGH-:
As a counter-example, systemd components check only stderr when deciding whether to upgrade to the native journal protocol or not: https://github.com/systemd/systemd/blob/40c05a34595ed769ce676206f3c5de874f9a9234/src/basic/log.c#L217-L241 (this is not a part of e.g. public libsystemd, it's used internally by various systemd binaries)

In theory, stdout and stderr  might be configured differently. `StandardOutput=`/ `StandardError=` can be changed explicitly, and shell-outs/shell pipelines are likely to make them different. It would be surprising if, for example, if stdout is configured to go to a file or a pipe, and application would redirect output (that it used to print to stdout) to the journal anyway (because stderr is still connected to the journal).

lucab:
The upgrade protocol is a bit subpar from the point of view of an application, as it uses a single env-variable and allow mixing two different concerns into that.
But I think you have a good point about mixed-use scenarios, and the docs also highlight the same case.
I'll reach out privately to the systemd folks for some customary feedback, but I guess it would be good to expose checks for stdout/stderr separately. That will give more flexibility to applications/consumers.

lucab:
I had an out-of-band chat with systemd upstream folks and I'm coming back with two main clarifications:
 * probing logic should be performed separately for stdout and stderr. That is, it makes sense to have two helpers to check `is *stdout* a journal stream?` and `is *stderr* a journal stream?` separately.
 * the protocol upgrading logic described in https://systemd.io/JOURNAL_NATIVE_PROTOCOL/#automatic-protocol-upgrading is only concerned with **stderr**. That is, if and only if stderr is a journal stream then automatic protocol upgrading is feasible.

lucab:
Given the above, the `StderrIsJournalStream()` method here is fine to stay.
Ideally it would be good to have a `StdoutIsJournalStream()` next to it.

Regarding protocol upgrade detection, this is effectively the same logic as the stderr check.
We can decide to either leave it out, or add something dedicated like `IsConnectedToJournal()` with a docstring to explain the protocol upgrade dance.
I have a slightly preference for the latter, as it helps capturing the nuances of this ticket discussion directly in the code.

journal/journal_unix.go
```diff
@@ -69,6 +69,44 @@ func Enabled() bool {
...
+// This can be used for automatic protocol upgrading described in [Journal Native Protocol].
+//
+// Returns true if JOURNAL_STREAM environment variable is present,
+// and stderr's device and inode numbers match it.
+//
+// Error is returned if unexpected error occurs: e.g. if JOURNAL_STREAM environment variable
+// is present, but malformed, fstat syscall fails, etc.
+//
+// [Journal Native Protocol]: https://systemd.io/JOURNAL_NATIVE_PROTOCOL/#automatic-protocol-upgrading
+func StderrIsJournalStream() (res bool, err error) {
```
lucab, line 84: I'd be happier if we could avoid named return values, as they can become the source of subtle bugs when capturing. Plain explicit return statements are better, thanks!

WGH-: Fixed

WGH- (commented):
I suppose I could try to write an integration tests that would run on CI. That would take some time, though. I'll try for a couple of days, and if I don't reply here, consider I've given up.

WGH-:
> Given the above, the `StderrIsJournalStream()` method here is fine to stay. Ideally it would be good to have a `StdoutIsJournalStream()` next to it.

So you want me to add it then?

> Regarding protocol upgrade detection, this is effectively the same logic as the stderr check. We can decide to either leave it out, or add something dedicated like `IsConnectedToJournal()` with a docstring to explain the protocol upgrade dance. I have a slightly preference for the latter, as it helps capturing the nuances of this ticket discussion directly in the code.

 `StderrIsJournalStream()`'s docstring already refers to https://systemd.io/JOURNAL_NATIVE_PROTOCOL/#automatic-protocol-upgrading, and there's an example of said dance that gets rendered in godoc, isn't it enough?

And if we introduce `IsConnectedToJournal()`, is it going to be simply an alias for `StderrIsJournalStream()`?

WGH-:
So I added `StdoutIsJournalStream`, and written some documentation as I saw fit. Please correct me if my understanding of its usefulness and limitations is wrong.

Additionally, I've written a somewhat complicated integration test for `StderrIsJournalStream` that executes itself as a transient systemd service to ensure it picks up `JOURNAL_STREAM` as it's initialized by systemd.

Unfortunately, it doesn't work as is on Ubuntu 16.04 due to [`systemd-run` not supporting `--wait`](https://github.com/coreos/go-systemd/actions/runs/3397443136/jobs/5649605794#step:14:133). Instead of adding more workarounds to an already complicated test, I bumped distro versions instead. Ubuntu 16.04 is old and not supported anyway (except for commercial ESM). If that's inappropriate, let's discuss other options.

lucab:
> I bumped distro versions instead. Ubuntu 16.04 is old and not supported anyway (except for commercial ESM). If that's inappropriate, let's discuss other options.

That's totally fine. The old version there is just because we hadn't recently touched the CI setup.

journal/journal_unix.go
```diff
@@ -69,6 +69,67 @@ func Enabled() bool {
...
+// [Journal Native Protocol]: https://systemd.io/JOURNAL_NATIVE_PROTOCOL/#automatic-protocol-upgrading
+func StderrIsJournalStream() (bool, error) {
+	return fdIsJournalStream(syscall.Stderr)
+}
+
+// StdoutIsJournalStream returns whether the process stdout is connected
+// to the Journal's stream transport.
+//
+// Returns true if JOURNAL_STREAM environment variable is present,
+// and stderr's device and inode numbers match it.
```
lucab, line 92: Typo: `stdout`

journal/journal_unix.go
```diff
@@ -69,6 +69,67 @@ func Enabled() bool {
...
+// is present, but malformed, fstat syscall fails, etc.
+//
+// This might report false negatives in certain configurations,
+// like when both stdout and stderr are independently connected to the
+// journal (the default is StandardError=inherit, which results in
+// both file descriptors getting the same duplicated socket, and doesn't
+// suffer from this problem):
+//
+//	StandardOutput=journal
+//	StandardError=journal
```
lucab, line 104: Interesting, I wasn't aware of this behavior. If you have a small reproducer for this, I have the feeling we could open a ticket for `systemd` and directly fix this annoyance there.

journal/journal_windows.go
```diff
@@ -33,3 +33,11 @@ func Enabled() bool {
 func Send(message string, priority Priority, vars map[string]string) error {
 	return errors.New("could not initialize socket to journald")
 }
+
+func StderrIsJournalStream() (bool, error) {
+	return false, nil
```
lucab, line 38: This and the one below should probably return an error stating that they are not implemented.

WGH-: My idea of how it's going to be used is as follows:
```go
ok, err := journal.StderrIsJournalStream()
if err != nil {
    fmt.Fprintln(os.Stderr, "StderrIsJournalStream: %v", err)
}

if ok {
    // configure logger to use journal native protocol
} else {
    // configure logger to use stderr
}
```

In this case returning an error would create a useless noise on Windows. Or library users should be careful to suppress this specific error on Windows, both complicating their own code, not to mention that (I suppose) many Go developers rarely test their code on Windows at all.

lucab: Fair point. I was expecting consumers to mostly merge together the `false` and `error` cases. I'm fine keeping this.

journal/journal_unix.go
lucab: Ack, thanks for the interesting feedback anyway.
I think the behavior with `journal+console` is not that much surprising, and also an uncommon case. To avoid adding cognitive load, I'd suggest dropping the warning note.

WGH-: Done

lucab (approved):
LGTM, thanks a lot for your contribution!

PR #414, opened 2023-01-03
workflows: update actions to current major versions

bgilbert:
Fixes [deprecation warnings](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/) for Node.js 12.


bgilbert:
Folded into #418.

PR #415, opened 2023-01-11
dbus: add helper methods for multiple start/stop with wait

pboyd04:
This allows for a single call to stop or start multiple units and wait until all the jobs are complete. This may seem a bit trivial, but I found myself needing these methods in a few places and rather than just wrapping this module with my own thought it might make sense to add to the upstream repo.


pboyd04:
Test failures are not due to this change.

kolyshkin (changes requested):
I think such functionality does not belong to this library, this is basic Go concurrency and there are many ways to implement it.

PR #417, opened 2023-02-08
workflows: add Go 1.20, switch to setup-go v4

bgilbert:
Cache dependencies and build outputs by default.


cgwalters (approved):

bgilbert:
Folded into #418.

PR #419, opened 2023-04-04
journal: fix return types inside C wrapper functions

darxriggs:
The wrapper functions already have the correct `void` return type. So the function pointers inside should use the same.

man pages
- https://www.freedesktop.org/software/systemd/man/sd_journal_open.html
- https://www.freedesktop.org/software/systemd/man/sd_journal_add_match.html


bgilbert:
Thanks for the PR!  Could you rebase onto current Git `main`?

darxriggs:
Sure. Done.

bgilbert (approved):

PR #421, opened 2023-05-17
sdjournal: add sd_journal_open_namespace wrapper

### LINKED ISSUE #413: sdjournal: is reading from journald namespaces supported?
I am using sdjournal indirectly with the filebeat journald input, and was wondering if reading from a non-default journald namespace is supported directly. I know that I can point filebeat to the namespace's systemd.journal file residing in `/var/lib/journal/machine-id.namespace`, but this requires knowledge of the machine-id for each individual server.
I solved the problem for our system by symlinking the `machine-id.namespace` directory to a machine-independent path, but this still feels hacked together.

Looking at the code it doesn't seem like, so I just wanted to confirm.

Thanks a lot and all the best!

igor-tsiglyar:
wrap sd_journal_open_namespace C call to enable reading journal records in non-default namespace

fixes #413

### DIFF
--- a/sdjournal/journal.go
+++ b/sdjournal/journal.go
@@ -39,6 +39,15 @@ package sdjournal
 // }
 //
 // int
+// my_sd_journal_open_namespace(void *f, sd_journal **ret, const char *namespace, int flags)
+// {
+//   int (*sd_journal_open_namespace)(sd_journal **, const char *, int);
+//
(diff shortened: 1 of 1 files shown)


igor-tsiglyar:
@lucab please review

igor-tsiglyar:
@bgilbert can you please review?

PR #426, opened 2023-07-19
add CancelJob()

tych0:
We'd like to be able to cancel jobs using this API, so add a wrapper for it.

### DIFF
--- a/dbus/methods.go
+++ b/dbus/methods.go
@@ -852,6 +852,11 @@ func (c *Conn) listJobsInternal(ctx context.Context) ([]JobStatus, error) {
 	return status, nil
 }
 
+// Cancel the specified job ID.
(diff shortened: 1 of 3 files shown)


tych0:
```
=== RUN   TestStartStopTransientUnitAll
    methods_test.go:1284: Test unit found in list, should be stopped
```

seems unrelated to this patch? but i don't see how it could happen in any case... `runStopUnit()` waits on the job status channel...

kolyshkin:
@tych0 if you're still interested, can you please rebase this? Also, you don't have to add `Context` suffix to the method name, it's only for those method that already have old (non-context) counterparts.

kolyshkin:
Closing in favor of #511.

PR #430, opened 2023-10-13
Bugfix/dbus methods context

j-licht:


kolyshkin:
@j-licht can you add test cases?

PR #431, opened 2023-11-03
dlopen: make error handling thread safe

Luap99:
The dlerror() value is stored in thread local storage and thus checking the error from go code is not safe because go can schedule goroutines on any other thread at any time. This means it is possible that between the dlsym() or dlclose() call the code get moved to another thread and thus the dlerror() value is not what we expect. Instead because we read now from another thread it will be an error from a different call.

A test is added which without this fix is able to reproduce the problem somewhat reliably.

This issue was observed in the podman CI[1] which uses the sdjournal API.

[1] https://github.com/containers/podman/issues/20569

### DIFF
--- a/internal/dlopen/dlopen.go
+++ b/internal/dlopen/dlopen.go
@@ -23,6 +23,7 @@ import "C"
 import (
 	"errors"
 	"fmt"
+	"runtime"
 	"unsafe"
 )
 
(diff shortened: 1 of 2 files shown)


Luap99:
Example failure with the new test but without the locked thread:
```
--- FAIL: TestDlopenThreadSafety (1.69s)
    dlopen_test.go:99: GetSymbolPointer strlen failed unexpectedly: error resolving symbol "strlen": libstrange3.so: cannot open shared object file: No such file or directory
    dlopen_test.go:99: GetSymbolPointer strlen failed unexpectedly: error resolving symbol "strlen": libstrange1.so: cannot open shared object file: No such file or directory
FAIL
FAIL    github.com/coreos/go-systemd/v22/internal/dlopen        1.751s
FAIL
```

cgwalters (approved):

jlebon:
Nice digging on this!

PR #441, opened 2024-05-04
Draft: Allow more control over socket-activated file descriptors

percivalalb:
The XY of the problem is the fds indices used by systemd socket activation are left empty once `activation.Listener` is called. Any new fds typically take the lowest available fds, which means the expected socket-activated indices may be populated by other fds at any other point in the executing of the program.

I'm creating a library that enables zero-downtime restarts by [`execve`'ing](https://man7.org/linux/man-pages/man2/execve.2.html) the same program which requires that the fds be populated in the expected indices. This is not possible if another part of the program is using those fds.

- This gives 3 methods of control, leaving currently logic intac: 
  - `ConsumeFiles` the default and matches the previous logic - closes the systemd fds passed in.
  - `ReserveFiles` "reserves" the systemd fds by atomically replacing them with fds pointing at `/dev/null`
  - `ConserveFiles` does nothing and keeps the systemd fds open 
- Also exposes the `unsetEnv` of the system

### DIFF
--- a/activation/filesmethod_unix.go
+++ b/activation/filesmethod_unix.go
@@ -1,0 +1,54 @@
+// Copyright 2015 CoreOS, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build !windows
+// +build !windows
+
+package activation
+
+import (
+	"fmt"
+	"os"
+	"syscall"
+)
+
+func (m Method) Apply(f *os.File) error {
+	saveFd := int(f.Fd()) // get the idx before being closed.
+
+	switch m {
+	case ConsumeFiles:
+		f.Close()
+	case ReserveFiles:
+		devNull, err := os.OpenFile(os.DevNull, os.O_RDWR, 0755)
+		if err != nil {
+			return fmt.Errorf("accessing /dev/null: %w", err)
+		}
+
+		nullFd := int(devNull.Fd())
+
+		// "If oldfd equals newfd, then dup3() fails with the error EINVAL."
+		if saveFd == nullFd {
+			syscall.CloseOnExec(nullFd)
+		} else {
+			// "makes newfd be the copy of oldfd, closing newfd first if necessary"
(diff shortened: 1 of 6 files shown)


activation/filesmethod_unix.go
```diff
@@ -0,0 +1,54 @@
...
+func (m Method) Apply(f *os.File) error {
+	saveFd := int(f.Fd()) // get the idx before being closed.
+
+	switch m {
+	case ConsumeFiles:
+		f.Close()
+	case ReserveFiles:
+		devNull, err := os.OpenFile(os.DevNull, os.O_RDWR, 0755)
+		if err != nil {
+			return fmt.Errorf("accessing /dev/null: %w", err)
```
kolyshkin: nit: no need to wrap an error as `os.Open` returns `PathError` which already has the file name.

activation/listeners.go
```diff
@@ -25,22 +25,29 @@ import (
...
+			}
 		}
 	}
 	return listeners, nil
 }
 
 // ListenersWithNames maps a listener name to a set of net.Listener instances.
-func ListenersWithNames() (map[string][]net.Listener, error) {
-	files := Files(true)
+func ListenersWithNames(opts ...option) (map[string][]net.Listener, error) {
```
kolyshkin: ditto

activation/listeners.go
```diff
@@ -60,8 +70,8 @@ func ListenersWithNames() (map[string][]net.Listener, error) {
 // TLSListeners returns a slice containing a net.listener for each matching TCP socket type
 // passed to this process.
 // It uses default Listeners func and forces TCP sockets handlers to use TLS based on tlsConfig.
-func TLSListeners(tlsConfig *tls.Config) ([]net.Listener, error) {
-	listeners, err := Listeners()
+func TLSListeners(tlsConfig *tls.Config, opts ...option) ([]net.Listener, error) {
```
kolyshkin: ditto

activation/listeners.go
```diff
@@ -81,8 +91,8 @@ func TLSListeners(tlsConfig *tls.Config) ([]net.Listener, error) {
 
 // TLSListenersWithNames maps a listener name to a net.Listener with
 // the associated TLS configuration.
-func TLSListenersWithNames(tlsConfig *tls.Config) (map[string][]net.Listener, error) {
-	listeners, err := ListenersWithNames()
+func TLSListenersWithNames(tlsConfig *tls.Config, opts ...option) (map[string][]net.Listener, error) {
```
kolyshkin: ditto

kolyshkin (changes requested):
Not sure if you still want to work on this. If yes, here are my initial comments.

percivalalb:
Hi, thank you for the review. I'll resolve the comments over the next week

percivalalb:
I've rebase onto main and added some more commits with changes based on your comments. I will squash the comments down once the changes have been reviewed.

Before deployment I will push some unit tests.

PR #444, opened 2024-06-25
journal: ability to set flags for journal instance

taranovegor:
I needed to merge journals, which involves using the `--merge` flag when calling `journalctl`. Since I couldn't find a way to do this in the current implementation, I propose the following solution.

The structure `JournalReaderConfig` is augmented with a field `Flags` that contains an array of constants. If the `Flags` array is not empty, the method `NewJournalWithFlags` is called with the bitwise OR of the values in the array.

```go
sdjournal.NewJournalReader(sdjournal.JournalReaderConfig{
    Since: time.Duration(-1) * time.Minute,
    Flags: []int{sdjournal.SD_JOURNAL_SYSTEM, sdjournal.SD_JOURNAL_CURRENT_USER},
})
```


kolyshkin (commented):
@taranovegor please let us know if you want to keep working on this.

taranovegor:
> @taranovegor please let us know if you want to keep working on this.

Yes, I will continue working on this, but I won’t be able to do it ASAP.

PR #446, opened 2024-08-23
daemon: SdNotify: add support for abstract unix sockets

arianvp:
As per https://www.freedesktop.org/software/systemd/man/latest/sd_notify.html#


WGH-:
I'm pretty sure Go has supported `@whatever` syntax out of the box since forever, so no explicit converison is necessary. Are you sure it doesn't work as it is?

arianvp:
I do not remember. This has been too long ago :)

kolyshkin:
> I'm pretty sure Go has supported `@whatever` syntax out of the box since forever, so no explicit converison is necessary. Are you sure it doesn't work as it is?

Indeed it should work as is (at least according to https://groups.google.com/g/golang-nuts/c/Swgdc-XBNi8?pli=1).

PR #454, opened 2025-01-06
journal: SendVals with arbitrary auto-converting functions

lzap:
Hello,

I am working on [log/slog](https://pkg.go.dev/log/slog) handler implementation for our project using this library and the API does only allow passing journald fields as a map of strings. The `log/slog` implementation is designed around lists of attributes (with some optimalization hacks to decrease allocation count), I would like to follow the suit and avoid conversion to maps.

This patch adds new function `SendVals` and `Value` interface which allows the API users to use alternative syntax to provide journald values:

```go
journal.SendVals(message, journal.PriInfo,
	journal.String("KEY", "VALUE"),
	journal.Int("INT_KEY", 1),
)
```

The whole suite of helper fucntions is provided, but they immediately convert the value to string since carrying the type does not make much sense in this case as the value is immediately written to byte buffer and sent to journald. It is just a syntactic sugar I thought might be useful, for my usecase only the `journal.String` is enough.


kolyshkin (commented):
I was looking at it and thought maybe we can somehow generalize it (to e.g. use `map[string]any`). For one thing, we can use fmt.Printf("%v") to serialize anything that go can serialize.

I _think_ go-dbus contains some code like this, need to take a look.

lzap:
Rebased.

> we can somehow generalize it

Not entirely sure what you mean by that, the current code already supports passing params as map via `Send(message string, priority Priority, vars map[string]string)` this adds an alternative function.

Luap99:
Inead of defining our own value interface would it make sense if we bump our go version requirement here to at least 1.21 which seems to be when slog was added and then just use slog.Attr 

That would mean we require slog but if the goal is to better integrate with that that seems to make sense to me, and then we actually get all the benefits of the type optimisations of the slog package,

lzap:
I am gonna be honest here, we ended up using the [newly-created official slog handler](https://github.com/systemd/slog-journal) under systemd umbrella. Therefore I would not be using this function, I do not like to write code that might turn to be useless, I am closing this PR.

However, if you really like this I can return back to it and do what you suggest, it is not a big change at all. Thanks!

PR #458, opened 2025-07-25
dbus: add AttachProcessesToUnit

kolyshkin:
The functionality is available since systemd v238 (see https://github.com/systemd/systemd/pull/8125/commits/6592b9759cae509b407a3b49603498468bf5d276) but was never exposed to go-systemd.

This is to be used by opencontainers/cgroups packages (see https://github.com/opencontainers/cgroups/pull/26) and eventually runc (see https://github.com/opencontainers/runc/pull/4822).

### DIFF
--- a/dbus/methods_test.go
+++ b/dbus/methods_test.go
@@ -1689,3 +1689,91 @@ func TestFreezer(t *testing.T) {
 
 	runStopUnit(t, conn, TrUnitProp{target, nil})
 }
+
+func testAttachProcessesToUnit(t *testing.T, subcgroup string) {
+	target := "attach-processes.service"
+	conn := setupConn(t)
+	defer conn.Close()
+
+	setupUnit(target, conn, t)
+	linkUnit(target, conn, t)
+
+	// Start the unit first.
+	reschan := make(chan string)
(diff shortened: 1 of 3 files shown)


dbus/methods_test.go
```diff
@@ -1689,3 +1689,90 @@ func TestFreezer(t *testing.T) {
...
+	if attachedUnit != target {
+		t.Fatalf("process was not attached to correct unit: got %s, want %s", attachedUnit, target)
+	}
+
+	// Cleanup
+	_, err = conn.StopUnit(target, "replace", reschan)
+	if err != nil {
+		t.Fatal(err)
+	}
+	<-reschan
```
Luap99, line 1769: Would it make sense to move this in a defer function after we start the unit, to stop the unit right away if a test above fails? I guess it is not a big deal to leak a unit with a sleep process.

kolyshkin: Yet better we can use t.Cleanup. Oops, not yet, in this repo we're stuck with a very old go version (for now).

Luap99 (approved):
LGTM

PR #459, opened 2025-07-29
Modernize and fix CI

kolyshkin:
Bump various versions, fix various issues, making CI working again. See individual commits for more details.

One particular thing -- branch protection rules now need to be modified to only require `all-done` jobs (see the last commit).

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -8,46 +8,22 @@ on:
 permissions:
   contents: read
 
-env:
-  # Minimum supported Go toolchain
-  ACTION_MINIMUM_TOOLCHAIN: "1.12.x"
-
 jobs:
   build:
     name: "Build"
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: ['1.17.x', '1.18.x', '1.19.x', '1.20.x']
+        go: ['1.12.x', '1.23.x', '1.24.x']
     steps:
       - run: sudo apt-get -qq update
       - name: Install libsystemd-dev
         run: sudo apt-get install libsystemd-dev
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Setup go
-        uses: actions/setup-go@v4
+        uses: actions/setup-go@v5
         with:
           go-version: ${{ matrix.go }}
-      - name: Go fmt
-        run: ./scripts/ci-runner.sh go_fmt
-      - name: Go build (source)
-        run: ./scripts/ci-runner.sh build_source
-      - name: Go build (tests)
-        run: ./scripts/ci-runner.sh build_tests
-      - name: Go vet
-        run: ./scripts/ci-runner.sh go_vet
-  build-minimum:
-    name: "Build on minimum supported toolchain"
-    runs-on: ubuntu-latest
-    steps:
-      - run: sudo apt-get -qq update
-      - name: Install libsystemd-dev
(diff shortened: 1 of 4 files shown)


cgwalters (approved):

.github/workflows/go.yml
```diff
@@ -8,24 +8,20 @@ on:
...
-  ACTION_MINIMUM_TOOLCHAIN: "1.12.x"
-
 jobs:
   build:
     name: "Build"
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: ['1.17.x', '1.18.x', '1.19.x', '1.20.x']
+        go: ['1.12.x', '1.24.x', '1.23.x']
```
Luap99, line 17: This is not in a chronological order which seems confusing.

go.mod
```diff
@@ -1,5 +1,5 @@
 module github.com/coreos/go-systemd/v22
 
-go 1.12
+go 1.22
```
Luap99, line 3: This seem unintentional. The commit message does not explain the bump and the test baseline you have set is still go 1.12?

internal/dlopen/dlopen_test.go
```diff
@@ -83,7 +83,7 @@ func TestDlopenThreadSafety(t *testing.T) {
 
 	// 10000 is the default golang thread limit, so adding more will likely fail
 	// but this number is enough to reproduce the issue most of the time for me.
-	count := 10000
+	count := 1000
```
Luap99, line 86: This is as well not document in any commit? IIRC I needed such a high number to reproduce the issue I fixed.

kolyshkin: My bad; this was not intended to be changed here. I experimented with this because locally it gave me a panic.

```
=== RUN   TestDlopenThreadSafety
runtime/cgo: pthread_create failed: Resource temporarily unavailable
```

(and then Go prints a huge panic message with almost 10000 goroutines).

Let's see if it works in CI now. If not, we need to address this somehow:

 - run this in a sub-process and check the result from the parent, considering `pthread_create failed` message as normal.
 - find another way to recover from the panic (currently I don't see a way);
 - find a way to assess how many goroutines can be run (it's complicated);
 - drop the test from the normal CI run;
 - make the count much smaller (on my machine, it succeeds with `count := 1000`, fails with `5000` and larger values).

go.mod
kolyshkin: This was unintended; removed.

.github/workflows/go.yml
kolyshkin: fixed

kolyshkin:
@Luap99 OK the current theory wrt TestDlopenThreadSafety panic is they've changed something in runtime so that the test is working fine when go 1.18 (and, presumably, newer) is used to compile it. It does not work with go 1.13 and go 1.15.

I'm going to test it.

Luap99:
@kolyshkin Ack, likely not worth to spend to much time trying to figure out what is going on these older versions. I am fine if the test gets skipped for now at least.

btw I invited you to this repo so you can update the merge protection for the new task names yourself.

kolyshkin:
OK, I've disabled TestDlopenThreadSafety for now (and opened https://github.com/coreos/go-systemd/issues/462 to not forget about it).

I've also removed old "required jobs" and added `all-done` ones being added here.

This PR is ready for review, can you take another look @Luap99?

PR #464, opened 2025-08-11
Bump go to 1.23, fix tests, add more images

kolyshkin:
Please see individual commits for details.

High level overview:
- move container setup logic to a script
- make apt-get install less verbose
- install known go version in containers
- go.mod: bump go to 1.23
- TestListJobs: fix flakiness
- login1: fix TestListSessions vs new systemd
- sdjournal: don't use fmt.Printf from tests
- djournal/TestJournalGetCatalog: fix vs new systemd
- add more images (newer ubuntu, debian, fedora)

These are moved to a followup PR:
- Replace interface{} with any
- Use for range for integers
- sdjournal/journal_test.go: use slices.Contains
- login1: remove unused method
- sdjournal,machine1: don't use rand.Seed in test
- Remove io/ioutil usage

### DIFF
--- a/.github/workflows/containers.yml
+++ b/.github/workflows/containers.yml
@@ -18,10 +18,11 @@ jobs:
       fail-fast: false
       matrix:
         baseimage: ['debian:bullseye', 'ubuntu:20.04', 'ubuntu:22.04']
+        go: [1.23, 1.24]
     steps:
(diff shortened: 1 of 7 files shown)


Luap99 (approved):
LGTM

PR #465, opened 2025-08-12
[test/DNM] add go 1.25

kolyshkin:


kolyshkin:
It works with go 1.25rc3 (coincidentally, go 1.25 is released today).

PR #466, opened 2025-08-13
Modernize sources, add golangci-lint

kolyshkin:
This can be split into a few major blocks:
1. Modernize source code to use facilities available in Go 1.22+.
2. Fix issues spotted by golangci-lint (in a minimal configuration).
3. Gofumpt the code.
4. Add golangci-lint to CI, remove old stuff (`gofmt` / `go vet`).

I tried hard to minimize the changes and be conservative, but might have failed in a few places.


unit/doc.go
```diff
@@ -0,0 +1,26 @@
+// Copyright 2015 CoreOS, Inc.
```
Luap99, line 1: What should do about the copyright header?
`Copyright 2015 CoreOS, Inc.` doesn't seem to make sense anymore.

Or just drop it entirely from the files? I don't think it is required, at least our other projects don't include it every file either.

unit/deserialize.go
```diff
@@ -38,10 +38,8 @@ const (
 	SYSTEMD_NEWLINE = "\r\n"
 )
 
-var (
-	// ErrLineTooLong gets returned when a line is too long for systemd to handle.
-	ErrLineTooLong = fmt.Errorf("line too long (max %d bytes)", SYSTEMD_LINE_MAX)
-)
+// ErrLineTooLong gets returned when a line is too long for systemd to handle.
+var ErrLineTooLong = fmt.Errorf("line too long (max %d bytes)", SYSTEMD_LINE_MAX)
```
Luap99, line 42: I don't mind that change but it seems unrelated to this commit.

dbus/methods.go
```diff
@@ -204,7 +205,8 @@ func (c *Conn) KillUnit(name string, signal int32) {
 // KillUnitContext takes the unit name and a UNIX signal number to send.
 // All of the unit's processes are killed.
 func (c *Conn) KillUnitContext(ctx context.Context, name string, signal int32) {
-	c.KillUnitWithTarget(ctx, name, All, signal)
+	// No way to return an error.
+	_ = c.KillUnitWithTarget(ctx, name, All, signal)
```
Luap99, line 209: I agree with the change but I wonder if we should mark this function as deprecated and let people use KillUnitWithTarget given this here just eats the error silently.

unit/deserialize.go
```diff
@@ -252,7 +252,7 @@ func (l *lexer) lexNextSectionOrOptionFunc(section string) lexStep {
 			return l.ignoreLineFunc(l.lexNextSectionOrOptionFunc(section)), nil
 		}
 
-		l.buf.UnreadRune()
+		_ = l.buf.UnreadRune()
```
Luap99, line 255: might be worth to a comment here saying that this should never error as we called ReadRune() before

Luap99 (commented):
LGTM mostly, only a few comments

unit/doc.go
kolyshkin: Yes, as far as I know there's no requirement to have a copyright in every source file. Will remove.

unit/deserialize.go
kolyshkin: My bad, it's gofumpt. Removed from this patch.

dbus/methods.go
kolyshkin: Was thinking the same thing. Amended.

kolyshkin:
@Luap99 addressed all your comments (and some of copilot ones).

Luap99 (approved):
LGTM

PR #468, opened 2025-08-18
ci: test go 1.25

Luap99:


Luap99:
> (we'll drop go 1.23 some time later I guess)

Yeah, it didn't saw any reason to drop 1.23 support right now. I rather not force upgrades unless we actually benefit.

PR #469, opened 2025-08-19
import1: test fixes

kolyshkin:
1. import1: use t.TempDir
    
    This test was leaving files and directories under $PWD.
    Use t.TempDir to avoid it.

2. import1: fix TestImport* cleanup
    
    TestImport functions create some artefacts under /var/lib/machines which
    require "machinectl remove" invocation to clean up.
    
    Unfortunately, Import operations are async, meaning we either have to
    listen to transfer events to ensure they are completed, or just sleep
    for some time. Since this is a test I'd like to keep things simple.

### DIFF
--- a/import1/dbus_test.go
+++ b/import1/dbus_test.go
@@ -18,14 +18,35 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 )
 
 const (
 	importPrefix = "importd-test-"
 )
+
+func cleanupImport(t *testing.T, name string) {
+	t.Cleanup(func() {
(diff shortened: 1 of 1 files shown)


import1/dbus_test.go
```diff
@@ -18,27 +18,46 @@ import (
...
+func cleanupImport(t *testing.T, name string) {
+	t.Cleanup(func() {
+		var (
+			out []byte
+			err error
+			dur = 500 * time.Millisecond
+		)
+		for range 5 {
+			time.Sleep(dur)
+			out, err = exec.Command("machinectl", "image-status", name).CombinedOutput()
```
Luap99, line 41: that is not removing anything, is it? Would this not need to call `remove`?

Issue #474, opened 2025-09-12
Race in dbus


purpleidea:
I seem to be able to trigger this fairly reliably. It pops up in two different ways, both are listed here.

I will admit that the code I wrote which uses this is very old and ugly, but a cursory look leads me to believe it's likely a bug in this library.

If you'd like more information, please let me know. FIgured I'd post this in case it's an obvious fix for the maintainers.

I'm happy to share code to reproduce it, but it's quite involved because https://github.com/purpleidea/mgmt/ which consumes this is not exactly small, so hopefully the trace is helpful.

Thanks!
```
==================
WARNING: DATA RACE
Read at 0x00c00235dc50 by goroutine 645:
  runtime.mapaccess2_faststr()
      /usr/lib/golang/src/internal/runtime/maps/runtime_faststr_swiss.go:162 +0x0
  github.com/coreos/go-systemd/v22/dbus.(*set).Contains()
      /home/james/code/gopath/pkg/mod/github.com/coreos/go-systemd/v22@v22.6.0/dbus/set.go:30 +0x68
  github.com/coreos/go-systemd/v22/dbus.(*SubscriptionSet).filter()
      /home/james/code/gopath/pkg/mod/github.com/coreos/go-systemd/v22@v22.6.0/dbus/subscription_set.go:29 +0x2b
  github.com/coreos/go-systemd/v22/dbus.(*SubscriptionSet).Subscribe.func1()
      /home/james/code/gopath/pkg/mod/github.com/coreos/go-systemd/v22@v22.6.0/dbus/subscription_set.go:38 +0xe
  github.com/coreos/go-systemd/v22/dbus.(*Conn).SubscribeUnitsCustom.func1()
      /home/james/code/gopath/pkg/mod/github.com/coreos/go-systemd/v22@v22.6.0/dbus/subscription.go:119 +0x318
  github.com/coreos/go-systemd/v22/dbus.(*Conn).ListUnits()
      /home/james/code/gopath/pkg/mod/github.com/coreos/go-systemd/v22@v22.6.0/dbus/methods.go:452 +0xd3
  github.com/coreos/go-systemd/v22/dbus.(*Conn).SubscribeUnitsCustom.func1()
      /home/james/code/gopath/pkg/mod/github.com/coreos/go-systemd/v22@v22.6.0/dbus/subscription.go:115 +0x9f

Previous write at 0x00c00235dc50 by goroutine 421:
  runtime.mapassign_faststr()
[...]

PR #480, opened 2025-10-24
subscription: Added context cancellation and sync to subscription set

NotSoFancyName:
Now it is possible to cancel unit subsciptions with context which eliminates existing goroutine leak
Added synchronization primitive for `SubscriptioSet`, now it should it be possible to add new units to the set after executing `SubscribeContext` 

Fixes: #474

### DIFF
--- a/dbus/subscription.go
+++ b/dbus/subscription.go
@@ -15,6 +15,7 @@
 package dbus
 
 import (
+	"context"
 	"errors"
 	"log"
 	"time"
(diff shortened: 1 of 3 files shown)


dbus/subscription.go
```diff
@@ -94,58 +95,77 @@ func (c *Conn) dispatch() {
...
 			}
+		} else {
+			errChan <- err
+		}
+	}
 
-			<-timerChan
+	go func() {
+		// Make the initial check for the unit statuses
+		sendSubscriptionUnitsUpdates()
```
Luap99: I don't follow why this was changed to an extra function. This makes a larger diff than necessary.

All you would need to chnage in the original code is use ListUnitsContext() and then where we wait for timerChan put the select which also waits for the Context and returns

NotSoFancyName:
Valid points

dbus/subscription.go
```diff
@@ -145,7 +156,12 @@ func (c *Conn) SubscribeUnitsCustom(interval time.Duration, buffer int, isChange
 				errChan <- err
 			}
 
-			<-timerChan
+			select {
```
Luap99: switch to ListUnitsContext on line 126 as well please

NotSoFancyName:
@kolyshkin 
Could you take a look?
Sorry for bothering you again

NotSoFancyName:
@kolyshkin 🙏 🥺

PR #484, opened 2025-12-05
activation: allow building without tls support

xnox:
When listeners.go APIs are not used by a given application, golang
doesn't automatically optimize them out. Add an opt-out build tag to
activation/listeners.go, such that one can build an application that
uses activation files without the listeners APIs.

For applications that do not already include any "crypto/tls" this can
be highly beneficial, for example runc can reduce it's total binary
size by 7.6%.

runc project already uses go mod vendor, and can just delete the listeners.go file:
- https://github.com/opencontainers/runc/pull/5056

But it would be preffered if this opt-out build-tag was available here
upstream for people to use in general.

### DIFF
--- a/activation/listeners.go
+++ b/activation/listeners.go
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build !no_listeners
+
 package activation
 
 import (


activation/listeners.go
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//go:build !no_listeners
```
Luap99: go build tags are not namespaced so there is always risk that it could conflict when another project uses the same name.

As such I would prefer something more like `go_systemd_no_listeners` or even `coreos_go_systemd_no_listeners`

Luap99 (commented):
I am fine with a build tag, but the listeners_test.go should use the same tag so it doesn't fail with the tag.

And I guess it would be nice to documented that tag in the Readme so all library users are aware.

cc @kolyshkin

Luap99:
Looks like something in the fedora CI tests broke, we need to look into that.

Otherwise I guess it would also be nice to add a CI test to make sure the build tag compiles correctly.

kolyshkin:
We've decided to solve the issue in runc in a different way (see https://github.com/opencontainers/runc/pull/5057), meaning this PR have no users now, thus closing.

kolyshkin:
> Looks like something in the fedora CI tests broke, we need to look into that.

Being fixed by https://github.com/coreos/go-systemd/pull/489

PR #485, opened 2025-12-07
unit: simplify escape character

huww98:


huww98:
@kolyshkin Done. thanks.

PR #493, opened 2026-01-07
dbus: dedup result conversion code

kolyshkin:
It looks like `listUnitsInternal` performs an unnecessary conversion from `[][]any` to `[]any`. Also, its code has been copied over a few times, with the only change to resulting type.

Introduce generic helper functions convertSlice and storeSlice to eliminate duplicated code. Also, remove unnecessary intermediate slice allocation and conversion.

This change:
- Adds convertSlice[T] to convert []any to typed slices
- Adds storeSlice[T] to fetch and convert D-Bus results
- Updates 9 functions to use the new helpers
- Removes ~150 lines of duplicated code
- Eliminates listUnitsInternal and listUnitFilesInternal wrappers

All functionality is preserved and tests pass.

Co-Authored-By: Claude Sonnet 4.5

### DIFF
--- a/dbus/methods.go
+++ b/dbus/methods.go
@@ -397,30 +397,33 @@ type UnitStatus struct {
 
 type storeFunc func(retvalues ...any) error
 
-func (c *Conn) listUnitsInternal(f storeFunc) ([]UnitStatus, error) {
-	result := make([][]any, 0)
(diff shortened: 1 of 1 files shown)


Luap99 (approved):
LGTM, not a huge fan of how go generics look but the deduplication does indeed look quite nice here.

PR #495, opened 2026-01-08
Add support for transient units with auxiliary units

### LINKED ISSUE #488: Expose auxiliary units in StartTransientUnit
The systemd dbus method `StartTransientUnit` takes a list of auxiliary units to start along with the main unit, this is used for things like starting a transient service with socket activation, by specifying the socket unit as the main unit and the service unit that is activated by it as an auxiliary unit.
The functions `StartTransientUnit` and `StartTransientUnitContext` on `dbus.Conn` in this package don't expose the auxiliary unit list and instead hardcode it to an empty list.

Would a PR to add this functionality be welcome? And if so, should it modify the existing `StartTransientUnit` and `StartTransientUnitContext` functions, or add a new function to keep backwards compatibility?

gwenya:
Closes #488

### DIFF
--- a/dbus/methods.go
+++ b/dbus/methods.go
@@ -194,7 +194,13 @@ func (c *Conn) StartTransientUnit(name string, mode string, properties []Propert
 // unique. mode is the same as in StartUnitContext, properties contains properties
 // of the unit.
 func (c *Conn) StartTransientUnitContext(ctx context.Context, name string, mode string, properties []Property, ch chan<- string) (int, error) {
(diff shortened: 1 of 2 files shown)


dbus/properties.go
```diff
@@ -39,7 +39,7 @@ type Property struct {
 	Value dbus.Variant
 }
 
-type PropertyCollection struct {
```
kolyshkin: Hmm, interesting, this is actually a breaking change (even if no one is using this type, it is part of public API).

The right way to maintain backward compatibility would be to add something like this:

```go
// Deprecated: please use [AuxiliaryUnit] instead.
type PropertyCollection = AuxiliaryUnit
```

PR #496, opened 2026-01-09
dbus: allow multiple calls for the same unit to *Unit

haircommander:
As well as add a test to prove StopUnit is reentrant

Alternative to https://github.com/kubernetes/kubernetes/pull/135826 
should fix https://github.com/kubernetes/kubernetes/issues/135825 (need to do some testing to double check)
credit to @rphillips for the idea and original piece of code

### DIFF
--- a/dbus/methods_test.go
+++ b/dbus/methods_test.go
@@ -1749,3 +1749,49 @@ func TestAttachProcessesToUnit(t *testing.T) {
 func TestAttachProcessesToUnitWithSubcgroup(t *testing.T) {
 	testAttachProcessesToUnit(t, "/test-subcgroup")
 }
+
+func TestStopUnitReentrant(t *testing.T) {
+	target := "start-stop.service"
+	conn := setupConn(t)
+
+	setupUnit(target, conn, t)
+	linkUnit(target, conn, t)
+
+	jobSize := len(conn.jobListener.jobs)
+
+	reschan := make(chan string)
+	// Buffered channels are important for multiple calls to the same job,
+	// so the order of when we pull them out doesn't need to matter.
+	reschan2 := make(chan string, 1)
+	reschan3 := make(chan string, 1)
+	errChan := make(chan error, 2)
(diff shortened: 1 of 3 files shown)


rphillips:
lgtm

haircommander:
@kolyshkin PTAL

dbus/methods_test.go
```diff
@@ -1749,3 +1749,49 @@ func TestAttachProcessesToUnit(t *testing.T) {
...
+	}()
+	go func() {
+		_, err = conn.StopUnit(target, "replace", reschan3)
+		errChan <- err
+	}()
+
+	<-reschan2
+	<-reschan3
+
+	for i := 0; i < 2; i++ {
```
kolyshkin: nit: `for range 2`

dbus/methods_test.go
```diff
@@ -1749,3 +1749,49 @@ func TestAttachProcessesToUnit(t *testing.T) {
...
+	}
+
+	<-reschan
+
+	go func() {
+		_, err = conn.StopUnit(target, "replace", reschan2)
+		errChan <- err
+	}()
+	go func() {
+		_, err = conn.StopUnit(target, "replace", reschan3)
```
kolyshkin: You use the same variable `err` in two goroutines. It's a data race:

| goroutine1| goroutine2
|------------- | ------------------
| err = StopUnit | |
| | err = StopUnit
| errChan <- err | |

I.e. the `err` can be rewritten by another goroutine before being send to the channel.

The fix is not to reuse err but rather instantiate a new one (in both goroutines):
```go
_, err := conn.StopUnit(...
```

kolyshkin (changes requested):
Left a few comments.

Also, if you can briefly describe the whole issue in the commit message that'd be awesome (in the ideal world I want to know why we're changing the code just by looking at `git log` or `git show`).

haircommander:
updated as suggested!

Luap99 (approved):
LGTM

PR #497, opened 2026-01-11
add FilesWithNames() to activation

MayCXC:
reopens https://github.com/coreos/go-systemd/pull/448 which had a broken actions runner.

### DIFF
--- a/activation/files.go
+++ b/activation/files.go
@@ -1,0 +1,29 @@
+// Copyright 2026 RedHat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


MayCXC:
@Luap99

activation/files.go
```diff
@@ -0,0 +1,29 @@
+// Copyright 2015 CoreOS, Inc.
```
kolyshkin: The copyright year should be 2025 since this code was written in 2025. Also, IANAL but the copyright holder should _probably_ be Red Hat, Inc. (as Red Hat acquired CoreOS in 2018).

kolyshkin:
If you can fix the copyright and squash your commits I think this one LGTM

kolyshkin (approved):

PR #498, opened 2026-01-14
activation: simplify ListenersWithNames

kolyshkin:
As map lookup returns nil if no key is found, and append will create a new slice if it's first argument is nil, we can greatly simplify the loop, as suggested in [1].

[1]: https://github.com/coreos/go-systemd/pull/448#discussion_r2676299340

### DIFF
--- a/activation/listeners.go
+++ b/activation/listeners.go
@@ -45,12 +45,7 @@ func ListenersWithNames() (map[string][]net.Listener, error) {
 
 	for _, f := range files {
 		if pc, err := net.FileListener(f); err == nil {
-			current, ok := listeners[f.Name()]
-			if !ok {
-				listeners[f.Name()] = []net.Listener{pc}
-			} else {
-				listeners[f.Name()] = append(current, pc)
-			}
+			listeners[f.Name()] = append(listeners[f.Name()], pc)
 			f.Close()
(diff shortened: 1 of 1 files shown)


Luap99:
@kolyshkin Did you only enabled rebase merges only and no longer the ones with merge commits?
Do we really need to force rebases all the time? Seems like low risk that something breaks because of that but I guess CI is fast enough here that the extra rebases don't matter much so I am fine with doing that if that was your intention.

kolyshkin:
> @kolyshkin Did you only enabled rebase merges only and no longer the ones with merge commits? Do we really need to force rebases all the time? Seems like low risk that something breaks because of that but I guess CI is fast enough here that the extra rebases don't matter much so I am fine with doing that if that was your intention.

I had this option enabled in branch protection rules:

> [ ] Require branches to be up to date before merging
> 
> This ensures pull requests targeting a matching branch have been tested with the latest code. This setting will not take effect unless at least one status check is enabled (see below).

but I've disabled it now. While in other repos I've seen a few cases where a PR caused a bug which was not caught before merging because of another recently merged PR, such cases are rare and we catch them anyway after the merge (and can do a quick fix), so it's probably not worth the hassle here.

PR #503, opened 2026-01-28
login1: Add NewWithConnection method 

invidian:
This method allows passing existing D-Bus connection, which allows to
re-use connection between clients and to mock D-Bus connection for
testing purposes.

Extracted from #390
Includes commits from #395

This PR replaces #396, since the original fork has been archived and I am not able to update it. CC @kolyshkin

### DIFF
--- a/login1/dbus.go
+++ b/login1/dbus.go
@@ -46,9 +46,7 @@ type Connection interface {
 	Object(string, dbus.ObjectPath) dbus.BusObject
 	Signal(ch chan<- *dbus.Signal)
 	Connected() bool
-	// TODO: This should be replaced with AddMatchSignal.
-	// See https://github.com/coreos/go-systemd/issues/388 for details.
-	BusObject() dbus.BusObject
+	AddMatchSignalContext(ctx context.Context, options ...dbus.MatchOption) error
 }
 
 // connectionManager explicitly wraps dependencies on established D-Bus connection.
@@ -394,14 +392,23 @@ func (c *Conn) Inhibit(what, who, why, mode string) (*os.File, error) {
 	return os.NewFile(uintptr(fd), "inhibit"), nil
 }
 
-// Subscribe to signals on the logind dbus
-func (c *Conn) Subscribe(members ...string) chan *dbus.Signal {
+// SubscribeWithContext subscribes to signals on the logind dbus. If adding match signals fails, an error is returned.
+func (c *Conn) SubscribeWithContext(ctx context.Context, members ...string) (chan *dbus.Signal, error) {
 	for _, member := range members {
-		c.conn.BusObject().Call("org.freedesktop.DBus.AddMatch", 0,
-			fmt.Sprintf("type='signal',interface='org.freedesktop.login1.Manager',member='%s'", member))
+		if err := c.conn.AddMatchSignalContext(ctx, dbus.WithMatchInterface(dbusManagerInterface), dbus.WithMatchMember(member)); err != nil {
+			return nil, fmt.Errorf("adding match for signal %s: %w", member, err)
+		}
 	}
 	ch := make(chan *dbus.Signal, 10)
 	c.conn.Signal(ch)
+	return ch, nil
+}
+
+// Subscribe subscribes to signals on the logind dbus. If adding match signals fails, errors are ignored.
+//
+// Deprecated: use SubscribeWithContext instead.
+func (c *Conn) Subscribe(members ...string) chan *dbus.Signal {
+	ch, _ := c.SubscribeWithContext(context.Background(), members...)
 	return ch
 }
 
--- a/login1/dbus.go
+++ b/login1/dbus.go
@@ -36,8 +36,35 @@ const (
 
 // Conn is a connection to systemds dbus endpoint.
 type Conn struct {
-	conn   *dbus.Conn
-	object dbus.BusObject
+	conn        Connection
+	connManager connectionManager
+	object      Caller
+}
+
+// Connection describes functionality required from a given D-Bus connection.
+type Connection interface {
+	Object(string, dbus.ObjectPath) dbus.BusObject
+	Signal(ch chan<- *dbus.Signal)
+	Connected() bool
+	// TODO: This should be replaced with AddMatchSignal.
+	// See https://github.com/coreos/go-systemd/issues/388 for details.
+	BusObject() dbus.BusObject
+}
+
+// connectionManager explicitly wraps dependencies on established D-Bus connection.
+type connectionManager interface {
+	Hello() error
+	Auth(authMethods []dbus.Auth) error
+	Close() error
+
+	Connection
+}
+
+// Caller describes required functionality from D-Bus object.
+type Caller interface {
+	// TODO: This method should eventually be removed, as it provides no context support.
+	Call(method string, flags dbus.Flags, args ...any) *dbus.Call
+	CallWithContext(ctx context.Context, method string, flags dbus.Flags, args ...any) *dbus.Call
 }
 
 // New establishes a connection to the system bus and authenticates.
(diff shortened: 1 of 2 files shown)


login1/dbus.go
```diff
@@ -51,14 +76,26 @@ func New() (*Conn, error) {
 	return c, nil
 }
 
+// NewWithConnection creates new login1 client using given D-Bus connection.
+func NewWithConnection(connection Connection) (*Conn, error) {
+	if connection == nil {
+		return nil, fmt.Errorf("no connection given")
```
kolyshkin: nit: use `errors.New`

kolyshkin (changes requested):
It looks like there are API changes, meaning we'll need to target this to v23.

For v23, we can probably remove some deprecated stuff as well, and even rename some functions (but I don't have a plan yet).

Left a few nits, too.

login1/dbus_test.go
```diff
@@ -21,18 +21,19 @@ import (
 	"regexp"
 	"testing"
 	"time"
+
+	"github.com/coreos/go-systemd/v22/login1"
```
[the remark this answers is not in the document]
kolyshkin: Either way is fine

kolyshkin (commented):
Perhaps you should reverse the order of the last two commits, unless there's a compelling reason to keep it the way it is.

Adding something with TODO in it and then immediately solving this in the next commit seems like an unnecessary cognitive load to a reviewer.

invidian:
If I swap the commits, I won't be able to test new code in [login1: Use AddMatchSignalContext instead of BusObject.Call directly](https://github.com/coreos/go-systemd/pull/503/commits/d01e98731f1a1fddfd638090aa14f847aa9f9f66), since [login1: Add NewWithConnection method](https://github.com/coreos/go-systemd/pull/503/commits/5508a35d415919305236cedafe8a6d33f24e9be8) enables testing, hence the original PRs split :)

login1/dbus.go
```diff
@@ -46,9 +46,7 @@ type Connection interface {
 	Object(string, dbus.ObjectPath) dbus.BusObject
 	Signal(ch chan<- *dbus.Signal)
 	Connected() bool
-	// TODO: This should be replaced with AddMatchSignal.
-	// See https://github.com/coreos/go-systemd/issues/388 for details.
-	BusObject() dbus.BusObject
+	AddMatchSignalContext(ctx context.Context, options ...dbus.MatchOption) error
```
kolyshkin, line 49: Since we don't have AddMatchSignal, there's no need to add `Context` suffix -- those are only for new functions that already had versions without context.

login1/dbus_test.go
```diff
@@ -245,9 +245,162 @@ func Test_Creating_new_connection_with_custom_connection(t *testing.T) {
...
+					return nil
+				},
+			}
+
+			conn, err := login1.NewWithConnection(connectionWithInterfaceCheck)
+			if err != nil {
+				t.Fatalf("Unexpected error creating connection: %v", err)
+			}
+
+			if _, err := conn.SubscribeWithContext(ctx, "SessionNew"); err != nil {
```
kolyshkin: Please use `t.Context()` instead. Same below.

kolyshkin (changes requested):
1. We've switch to Go 1.24 as a minimum version and thus can (and should) use `t.Context()` in tests.
2. A naming nit.

Otherwise LGTM

login1/dbus.go
invidian: Code depending on `Connection` interface is expected to accept https://pkg.go.dev/github.com/godbus/dbus/v5#Conn.AddMatchSignalContext, so the methods must match this struct.

kolyshkin (approved):
LGTM

@Luap99 PTAL

Luap99 (changes requested):
I hate interfaces as part of stable APIs, it basically comes down to the fact that there is no way to even add methods without breaking users of the interface.

We will also not be able to switch any methods that we use on *dbus.Conn without breaking people using NewWithConnection() as I consider this not useful.

yes it is useful for mocking but only that?! as such it seems not worth for me to burden the stable API for a bit of tests, for testing these things should be private types.

For the public API if this is only about connection reuse then I guess the answer should be accept `*dbus.Conn`?

Also what makes login1 special, do we not have the exact same pattern in dbus, import1, machine1 as well? So it seems odd only allowing this reuse in one package?

invidian:
Thanks for the review and comment @Luap99. This is an old PR, so I forgot the context a bit as well, but reading original issues and PRs, the reasoning in the PR description stands accurate, although could be extended.

Right now every package (login1, machine1 etc.) hardcodes and repeats the authentication code used for establishing the connection, which:
- might be source of bugs, since if one implementation has bug and is fixed, it is not guaranteed other implementations will receive the same fix.
- users are not able to customize how the authentication is done, limiting the library adoption
- dbus connection cannot be reused across clients, which wastes resources

IMO accepting a connection is a good step to improve all these, while keeping code backward compatible.

> ...
> For the public API if this is only about connection reuse then I guess the answer should be accept *dbus.Conn?

I think your concern about the interface is right. However, this is the exact purpose of using the interface, to create an boundary, which is more stable than internal details of code and testable code should accept its' dependencies via interface, as e.g. google style guide says via [Accept interfaces, return concrete types](https://google.github.io/styleguide/go/best-practices). In addition, accepting an interface instead of `*dbus.Conn` specific type guards this piece of code from the changes in `*dbus.Conn` _and_ makes the library easier to test for consumers, where they can stub login1 code when testing, but use actual client in production code.

Maybe the interface is wrong or too big in this particular case and some other wrapper should be used? Would that help?

> Also what makes login1 special, do we not have the exact same pattern in dbus, import1, machine1 as well? So it seems odd only allowing this reuse in one package?

Login1 is not special, it's just the specific code I care about as part of https://github.com/flatcar/flatcar-linux-update-operator project, which suffers from original issue right now. This is just setting a pattern which should probably be extended to all packages, if accepted, although that might be out of scope for this particular PR?

Luap99:
> I think your concern about the interface is right. However, this is the exact purpose of using the interface, to create an boundary, which is more stable than internal details of code and testable code should accept its' dependencies via interface, as e.g. google style guide says via [Accept interfaces, return concrete types](https://google.github.io/styleguide/go/best-practices). In addition, accepting an interface instead of *dbus.Conn specific type guards this piece of code from the changes in *dbus.Conn and makes the library easier to test for consumers, where they can stub login1 code when testing, but use actual client in production code.

> Maybe the interface is wrong or too big in this particular case and some other wrapper should be used? Would that help?

Well my point is there is zero guarantee that interface is stable, in fact you decided to implement in in commit 2 and then break it right away in commit 3 clearly signalling me that is not a stable base we should use in a public API.
A interface like that locks us into the few exposed dbus.Conn methods with no way of updating if better methods become available.

IMO the only reasonable thing is to accept the same API commitment as github.com/godbus/dbus/v5 which is dbus.Conn. I understand that this will not allow mocking of tests but I don't think we should leak that into the public API just for tests.

invidian:
Let me see how consumer (FLUO) code would look like with change like this.

invidian:
With function accepting `*dbus.Conn`, it does improve FLUO code still and things remain testable as long as login1 code is placed behind an interface.

But with most of tests stripped from this PR, due to not being able to test more, this PR would boil down to pretty much just:
```diff
+// NewWithConnection creates new login1 client using given D-Bus connection.
+func NewWithConnection(connection *dbus.Conn) (*Conn, error) {
+       if connection == nil {
+               return nil, errors.New("no connection given")
+       }
+
+       return &Conn{
+               conn:   connection,
+               object: connection.Object(dbusDest, dbusPath),
+       }, nil
+}
```

Which perhaps should be added across all packages, not just to login1, which derives a lot from my original idea for this contribution. On top of that, I no longer actively maintain FLUO, so it is not that important overall to get this PR merged.

I'd be happy to drive this work to at least get `NewWithConnection` in, either in just login1 or all packages, but I'm fine with just closing this PR too at this point.

PR #514, opened 2026-03-11
dbus: Don't deadlock on shutdown

purpleidea:
When cancelling the ctx, if we have a pending channel send, then we'll deadlock. The correct pattern is to abort if ctx exits.

We also close the channels when the sender closes.

### DIFF
--- a/dbus/subscription.go
+++ b/dbus/subscription.go
@@ -120,6 +120,8 @@ func (c *Conn) SubscribeUnitsCustomContext(ctx context.Context, interval time.Du
 	errChan := make(chan error, buffer)
 
 	go func() {
+		defer close(statusChan)
+		defer close(errChan)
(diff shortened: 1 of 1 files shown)


purpleidea:
Not sure how I didn't notice this deadlock sooner. I think it's important to get this in somehow.

Luap99 (commented):
Thanks, just one minor thing. Also CI just get fixed by https://github.com/coreos/go-systemd/pull/515

purpleidea:
> Thanks, just one minor thing

Not a minor thing, I just plainly was coding fast and not paying attention. All fixed and rebased. Cheers

Luap99 (approved):
LGTM

kolyshkin (approved):
LGTM

PR #515, opened 2026-03-13
Upgrade to go 1.24 and test with go 1.26

Luap99:
see commits


kolyshkin (approved):
Just a single nit, otherwise lgtm

PR #513, opened 2026-03-06
update golangci-lint to v2.10.1

Luap99:
The older version does not support go 1.26 but the workflow installs go 1.26 as it ask for the stable go version. Therefore CI fails currently. Fix this by updating golangci-lint to the latest version which does support go 1.26.

PR #516, opened 2026-03-14
scripts/ci-runner.sh: wait for systemd to settle

kolyshkin:
The sleep stage, originally added by commit 07d38cb, is there to wait for systemd to finish its initialization stage (which includes wiping /tmp).

Commit 743c5ad added a step of building tests inside a container, and then commit 82dc44e moved the code into scripts/ci-runner.sh.

Now, sometimes building the tests fails with errors like this:

> github.com/coreos/go-systemd/v22/activation.test: open /tmp/go-build4063755751/b001/importcfg: no such file or directory

or

> asm: open $WORK/b006/symabis: no such file or directory

(as reported in https://github.com/coreos/go-systemd/pull/513 comments).

All this points to /tmp being wiped while the tests are being built.

Moving the sleep to before building tests (right after starting the container) should fix this. Yet better, let's use

	systemctl is-system-running --wait

to wait for systemd to settle.

### DIFF
--- a/scripts/ci-runner.sh
+++ b/scripts/ci-runner.sh
@@ -65,10 +65,8 @@ EOF
     docker run --shm-size=2gb -d --cidfile="$cidfile" --privileged -v "${PWD}:/src" "$name" /sbin/init --system
     cid=$(cat "$cidfile")
     rm -f "$cidfile"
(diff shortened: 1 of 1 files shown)


Luap99 (approved):
LGTM

PR #522, opened 2026-07-20
Fix handle leak

ProjectMutilation:
Add missed Close() and removing in error path
Fixes: ca20d25 ("journal: split into read and write packages")
Found by PostgresPro


journal/journal_unix.go
```diff
@@ -243,6 +243,8 @@ func tempFd() (*os.File, error) {
 	}
 	err = syscall.Unlink(file.Name())
 	if err != nil {
+		file.Close()
+		os.Remove(file.Name())
```
Luap99: there i no point in calling os.Remove() as unlink already failed so calling unlink again here here seems unlikely to ever work.

Luap99 (changes requested):
> Fixes: da560ea ("Build on Windows")

If you include a fixes line it must produce the actual commit which introduced the problem. This commit just moved the code elsewhere so listing this is is just incorrect and misleading.

journal/journal_unix.go
ProjectMutilation: you are right, I'll fix both remarks

PR #527, opened 2026-08-17
dlopen: report the dlerror from GetHandle

vtushar06:
I ran into this from the podman side - the API returns `unable to open a handle to the library` and there is no way to tell why libsystemd did not load. `GetSymbolPointer` and `Close` have reported `dlerror()` since #431, `GetHandle` is the one that never calls it.

The case I care about is when the file is present and dlopen still refuses, so "not found" is misleading:

```
old:  unable to open a handle to the library
new:  unable to open a handle to the library: /tmp/libbroken.so.0: invalid ELF header; libstrange.so: cannot open shared object file: No such file or directory
```

`ErrSoNotFound` is still wrapped so `errors.Is` keeps working, and the bare error is kept when dlopen gives no reason. `runtime.LockOSThread` for the same reason as #431.

The test fails on current main:

```
--- FAIL: TestGetHandleError (0.00s)
    dlopen_test.go:82: error does not mention "libstrange1.so": unable to open a handle to the library
    dlopen_test.go:82: error does not mention "libstrange2.so": unable to open a handle to the library
```

Context, if useful: https://github.com/podman-container-tools/podman/pull/29528


amccabe:
I hit this failure outside of CI, and reported issue podman-container-tools/podman#29822 for it (I'd opened #528 with the same change as this PR, I defer to you and noted it in the issue comments).

